PR #2, opened 2022-02-24
Illustrate how to use markdown notebooks as blog posts

stefanv:


stefanv:
@tupui @rossbar I don't quite know how to structure this PR.

I currently have a script that takes `notebook.md` (a Jupytext notebook), then executes it to produce `index.md` which Hugo can render.

Do you think this is something the blog author would need to do manually each time they want to update their post?  I don't think we want to execute all blog posts every time we build the site (an unnecessary CPU churn, and also not clear the blog posts would even execute after a few years, e.g.).

So, how do I make this easiest for authors?  Simply documenting that they run `nb2blog` every time they've updated the markdown file?

And, of course, I can make nb2blog friendlier so that it does some argument checking, lives in a different directory, etc.

rossbar:
My vote would also be for cached builds, i.e. only re-executing posts that are new or changed. As far as UX building/previewing blogposts, I don't have any particular thoughts other than it should be simple and able to run locally; e.g. a command like `make html` to execute+convert posts and `hugo serve` to preview the results. As long as the workflow is simple, clear, and works both locally and in CI w/ preview, then I think it can be structured however!

stefanv:
OK, this is ready to go.  For now, it installs the dependencies necessary to build everything on Netlify.  Eventually we should remove that capability, since the blog posts should be compiled offline—not by Netlify.

jarrodmillman (approved):

PR #17, opened 2022-03-12
Add matplotlib content

stefanv:
@tupui I was wondering if you'd be willing to take a look and perhaps take this one further.

- The featuredImage should maybe be rendered as part of the summary text, not the post header (not sure what would look best).  As it is it looks okay, except that for very short post summaries the next post header overlaps with the image.
- This is about 100 MB of content for 38 blog posts; we may want to think about what we want to do about repo size in the long run.
- Not all these blog posts may be relevant, such as the introductory one, as well as the one that discusses how to make a blog post (that to be moved to README?)

I changed the font of the site to serif; I think that reads a lot better.

Let me know your thoughts!


jarrodmillman:
@stefanv Would you rebase on main?  I am not sure, it will solve the problem; but I did fix a build issue earlier today.  I also updated a few other things.

tupui:
Yes would be happy to have a look. Next week I can take some time for it.

stefanv:
@jarrodmillman The blog now renders fine, but it does not pass the thresholds for the quality checker. Is it OK to reduce those for now, or should we dig into the build to see what can be improved?

stefanv:
Remaining lighthouse errors:

```
4:24:53 PM: Expected category Accessibility to be greater or equal to 0.95 but got 0.84
4:24:53 PM:    'Background and foreground colors do not have a sufficient contrast ratio.' received a score of 0
4:24:53 PM:    '`[id]` attributes on active, focusable elements are not unique' received a score of 0
4:24:53 PM:    'Image elements do not have `[alt]` attributes' received a score of 0
4:24:53 PM:    'Links do not have a discernible name' received a score of 0
4:24:53 PM: Expected category SEO to be greater or equal to 0.95 but got 0.92
4:24:53 PM:    'Image elements do not have `[alt]` attributes' received a score of 0
```

stefanv:
See also https://github.com/scientific-python/scientific-python-hugo-theme/pull/153

rossbar:
I was confused by something just looking at the diff - is the addition of all the static images (pngs + gifs) intentional? I thought the blog posts were to be executed and would've expected the resulting images to be generated during execution rather than saved and uploaded as part of the repo.

tupui:
We discussed a bit about the architecture yesterday and we are not sure yet. For now we will make things simple to make it roll, but we have a few options to evaluate later. Ex. a repo per year, an external storage space, a repo for the site and a repo for submission, etc.

jarrodmillman:
@tupui I already spoke to @stefanv about this. I am going to pull some of the non-matplotlib specific stuff out of this PR.  I will put them in separate PRs, merge them, and rebase / redo this one.  Let me know if that is going to cause you problems.  I have some time to do it now, if that's ok.

tupui:
Ok I had some time to explore. This is fire! 👏 I have the following points for now (note that this is in general and not related specifically to the content):

* I like the titles, thanks for not underlying!!!! I hate that in some blogs, bold was made for that.
* I am not sure about the navbar. Is it intended that we redirect to learn.xxx? So all the information about how to propose something should be there? And `Community Blog` is redundant. In the end I think the blog could be part of the main website. Cf the Spotify example bellow, the blog is integrated in the website. Then we would need to present in a sidebar for ex. how to contribute.
* We should not load all articles at once. Not a fan of pages, but something like infinite scroll automatic refresh would be good or a `Read more` button.
* We need a way to select tags clearly. Ex [Spotify](https://research.atspotify.com/blog/) blogs has a filter dropdown
* About comments. I personally don't like comments: (i) needs a lot of moderation if this takes off, (ii) would just ask readers to post on the forum. I would not create automatically a post because I expect most article to be empty. This would pollute the forum.
* There is an overlap of some articles. See the screenshot.
* A serif font is fine although I personally prefer a sans serif one for blogs. It looks more "tech"
* We need a dark mode button.
* When we have notebooks, we should have a clear button on the side to download it. We also need a copy button on the code cells.
* For the content itself, yes we should remove the intro post and the how to contribute one.

Screenshot 2022-03-15 at 21 09 34

tupui:
* Need some padding between the image and the info-image.
Screenshot 2022-03-15 at 21 37 21Screenshot 2022-03-15 at 21 38 15

tupui:
* For the author's name, it would be good to have a contributor page. Since we want to open this up to writer which are not necessary on GitHub, I think we cannot go with just the GH handle. I would suggest to make a file with contact info that a user provide. Then we can link these and you could filter by author. Clicking on a author name, the page would start with a bio, contact info, pics (what was provided) and then the articles. This could be the same mechanism for the tags, we just need to consider tags as authors too for that feature. So tags would have a description too.

jarrodmillman:
> I was confused by something just looking at the diff - is the addition of all the static images (pngs + gifs) intentional? I thought the blog posts were to be executed and would've expected the resulting images to be generated during execution rather than saved and uploaded as part of the repo.

This is just a port of the existing matplotlib blog, which doesn't use notebooks.  I don't think it is worth going back over old content.  The notebook option is for authors who want to use it.  So I suspect in the future, some authors will use it and some won't.

stefanv:
> > I was confused by something just looking at the diff - is the addition of all the static images (pngs + gifs) intentional? I thought the blog posts were to be executed and would've expected the resulting images to be generated during execution rather than saved and uploaded as part of the repo.
> 
> This is just a port of the existing matplotlib blog, which doesn't use notebooks (as I recall). I don't think it is worth going back over old content (we have plenty of new work to do). The notebook option is for authors who want to use it. So I suspect in the future, some authors will use it and some won't.

Even for the notebooks we agreed to commit the rendered product; we don't want to be re-generating the output for each blog-post each time we compile the website.  Besides, content may get outdated, and we don't really care.  This isn't like documentation that needs to stay working.

stefanv:
- Infinite scrolling in Hugo: https://mikeroibu.com/post/hugo-infinite-scrolling/
- Dark mode: easiest would be to recommend https://darkreader.org/, otherwise [set color vars](https://www.w3schools.com/Css/css3_variables_javascript.asp) on click of a button
- Filter on multiple tags: https://github.com/pointyfar/hugo-tags-filter

The rest should be straightforward Hugo / CSS.

I changed the font from sans to serif; I think it helps with long-form reading. Medium does the same, see e.g. https://medium.com/@levmaximov  But, if you find another sans font that's nice and readable, go for it.

tupui:
Found this for dark mode https://yonkov.github.io/post/add-dark-mode-toggle-to-hugo/

I will do some tests for the font. I will let you know if I find something.

stefanv:
@tupui See https://github.com/scientific-python/scientific-python-hugo-theme/pull/157 for a proof-of-concept dark-mode switcher

PR #41, opened 2022-03-24
Update governance document

jarrodmillman:


content/about/governance.md
```diff
@@ -1,3 +1,79 @@
 ---
 title: "Governance and Decision Making"
 ---
+
+{{< notice warning >}}
+This is a draft document.
+{{< /notice >}}
```
tupui, line 7: I would remove this, the document looks good enough IMO

tupui (approved):
It reads good and renders good, thanks! Just noted we don't yet have the reviewers team.

tupui (approved):

PR #52, opened 2022-04-08
Configure mathjax

jarrodmillman:
We were using a retired CDN and mathjax 2.7:
- https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML
- https://www.mathjax.org/cdn-shutting-down

I followed the instructions here:

- https://docs.mathjax.org/en/latest/web/configuration.html
- https://docs.mathjax.org/en/latest/options/input/tex.html

We can also configure the output.  For example,
- https://docs.mathjax.org/en/latest/options/input/tex.html

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -2,10 +2,9 @@
 MathJax = {
   tex: {
     inlineMath: [['$', '$'], ['\\(', '\\)']],
-  },
-  loader: {load: ['ui/lazy']}
+  }
 };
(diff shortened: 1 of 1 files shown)


tupui (approved):
LGTM otherwise 👍 CDN is the recommended one and config looks fine.

tupui:
Nice performance bump! From 75 to 90

PR #58, opened 2022-04-26
NumPy's random number generators

tupui:
Objective is to present the following:

1. How bad it is to seed: example with setting a classical 123456 and showing the resulting distribution, what you get with this and explain that you don’t have random numbers if you set a seed. I would redirect to real deterministic methods like QMC for such use cases.
2. Why global state is wrong: maybe an example with multithreading being not reproducible.
3. How to use Generator
4. How to test with random numbers: RandomState + seed (as per NEP19)


tupui:
Let's make a call tomorrow 😃 Reproducibility aside from testing is something that I actually fight for RNG. sure I am biased since I am from the uncertainty community and have a background in QMC.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Welcome to `np.random.Generator`
+
+Now you might have read that `np.random.RandomState` is deprecated and
+`np.random.Generator` should be used instead. It's true, but a bit more nuanced
+as we will see. What is true though is that in any case, a global state
+generator must not be used.
+
+So why this new interface `np.random.Generator`? The main reason is to break
+away from backward compatibilities concerns. `np.random.RandomState` was
```
rgommers, line 132: You may actually want to say something about the split between `Generator` and `BitGenerator` instances, and the ability to provide multiple algorithms (and that NumPy actually now _does have_ state of the art algorithms). And parallel capabilities, and performance.

rgommers (commented):
A few drive-by comments on the new API. It's also important to make explicit who your audience is here. I think it is "end users" or "authors of research code". A blanket "this is wrong" feels bad otherwise, because it's definitely not always wrong - that's way too black-and-white. Tests suites tend to have hundreds of `np.random.seed` calls in them for example, and in practice that's okay for reproducible unit tests (and probably preferable over much more verbose use of high-entropy seeds).

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Random done wrong
+
+```python
+# the following is wrong and only exists for demonstration purposes!
+np.random.seed(123456)
+sample = np.random.random(1_024)
+```
+
+Aoutch! Two lines of codes and both are wrong. This is a simple snippet to
```
seberg, line 40: Strong word, "wrong" :).  I don't mind it too much, I just tend to try to be a bit less strong.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+
+# [0.33622174]
+# [0.5430262]
+# [0.96671784]
+# [0.37674972]
+# [0.96671784]
+```
+
+As you can see, although we started with the same state, this program is not
+deterministic anymore. This is what we mean by thread safety.
```
seberg, line 84: You have not introduced "thread safety", maybe rephrase slightly to say, that this means this is not "thread safe"?  (To read more like a definition of the term, rather than like it clarifies something introduced previously)

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+It sets the global seed and defines a `rng` using the object based API.
+
+> RNG stands for Random Number Generator. This abbreviation is commonly used.
+
+Then `np.random.random` and `np.random.RandomState` calls 5 times. The global
+state is getting updated, but it does not impact our `rng` stream of numbers.
+
+A lot of packages already migrated their code to move away from the global
+generator in favour of an object one. In this case, setting the global state
+would have absolutely no effect and you will end up in a broken state.
```
seberg, line 122: > you will end up in a broken state.

I am not sure what is meant?  But I also wonder if you need a longer setion on `RandomState` at all.  After all, we want users to use `Generate`, so new users have less names flying around in their heads.
(OTOH, I agree that it is a useful on-ramp to the global state problem, so just thought.)

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+[here]({{< relref "../../scipy/qmc-basics" >}})
+).
+
+If you want or need to use NumPy for that, then either use
+`np.random.Generator` or the legacy `np.random.RandomState`. Aside from
+backward compatibility concerns, `np.random.RandomState` is still present in
+NumPy for testing.
+
+```python
+rng = np.random.default_rng()
```
seberg, line 259: `default_rng()` is not necessarily stable.  So I think for testing `RandomState` should be used?

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+    axs[i].legend()
+```
+
+![
+Histograms for uniform, normal and gamma distribution using 1024 samples.
+They are compared with the underlying distributions and a poor match. 
+](distributions_seeded.png)
+
+Does that look good to you? It is not. If you need to have a determined
+sequence of points, there are other technics that would generate almost
```
seberg, line 215:
```suggestion
sequence of points, there are other techniques that would generate almost
```
I think?

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+perfect histograms only using a fraction of the number of points which is used
+here. See [this article]({{< relref "../../scipy/qmc-basics" >}})
+for more information on this.
+
+But why does it matter? How bad this is will depend on your application.
+If you don't, statistical properties are lost. This is hard to get because for
```
seberg, line 221:
```suggestion
If you don't, statistical properties are lost. This is hard to understand because for
```

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+most people, random is random. So taking an arbitrary draw of sample should not
+matter. And it does not **if** you replicate your analysis using different
+**random** seeds. This is the part everyone is forgetting. Whether you are
+optimizing something or calculating some statistics, you could get (un)lucky
+if you do this just once.
+
+If your application needs random numbers, then do use random numbers and play
+with the fact that they are random. You do science or anything critical
+using these randomly generated numbers, then please stay away from seeds.
```
seberg, line 230: I am wondering if we shouldn't say that we should use seeds carefully or only when reproducibility becomes necessary.  Also validating that your seed did not happen to lead to non-ordinary results is important.

I am alergic to the word "people" myself, so I would rephrase the next paragraph slightly into "we/I often see code where the seed was copy pasted from elsewhere without much thought"...  But this may lead to biases when possibly the same pseudo random numbers are used for similar analysis.  Never changing the seed also fails to check whether the analysis is robust.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+together to stay healthy. As a whole, a given forest will have a sort of
+homogeneity, a character. Now, burn everything down and use a single seed to
+rebuild your forest. This results in a monoculture, which is poor and less
+robust to the environment.
+
+Setting the seed of a random number generator is just like that. There is
+nothing random anymore, you are breaking your random number generator, breaking
+your forest. You see often in **code** the exact same seeds: 0, 123456.
+In fact, if you do this, you are using a fixed sequence of points which
+effectively has no guarantee over the quality of the points.
```
seberg, line 176: Do we like funny half fitting references like this one: https://dilbert.com/strip/2001-10-25

The one tree we get is still just as random, but the forest will not look random anymore?  Maybe we should stress it a bit, that the tree is still pseudo random, but the forest may lose some of its randomness?

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+
+But because you can use a seed doesn't mean you can resort to our old friends
+0, 123456, etc. NumPy proposes a canonical way to randomly give you a seed.
+
+```python
+import numpy as np
+print(np.random.SeedSequence().entropy)
+```
+
+Use this snippet to generate seed values for your tests.
```
seberg, line 284: The `SeedSequence` was (as I understand it) a milestone in conveniently creating multi-threaded seeded streams, and seeding in general.  I wonder if you want to talk about that a bit more?

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+
+[NEP19](https://numpy.org/neps/nep-0019-rng-policy.html) actually
+recommends `np.random.RandomState` for testing as the byte stream is not
+guaranteed as strongly as before with the new generators. In practice, changes
+are not happening often though. What matters in both cases is that you use the
+object based API and not the legacy approach with global state.
+
+Now, one thing with tests is that we want them to be reproducible so that
+we can inspect our code and find issues. Here you want to use a seed.
+This is the **only** valid use for a seed. (Sure if you are an expert in
```
seberg, line 272: Reproducibility is.  But reproducibility is not confined to testing.  My gut feeling would not be to paint seeding as dark here.  But rather say clearly that adding the seeding is best done only when reproducibility becomes vital.  I.e. not, if the analysis fails half the time otherwise...

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+    axs[i].plot(x, dist.pdf.pdf(x), 'r-', lw=5, alpha=0.6, label=dist.name)
+    axs[i].legend()
+```
+
+![
+Histograms for uniform, normal and gamma distribution using 1024 samples.
+They are compared with the underlying distributions and a poor match. 
+](distributions_seeded.png)
+
+Does that look good to you? It is not. If you need to have a determined
```
seberg, line 214: Can you explain why it is bad?  I am honestly not sure, the paragraph is about seeding, but this should not be bad due to seeding?  Certainly, the RNGs are a perfectly valid sample for that distribution?
Is this an advertisement for using better ways to generate random numbers for certain use-cases, or a case against seeding itself?

content/posts/numpy/generator-rng/index.md
seberg: That would be nice, although if the main part here is about the global state and the seeding, it may also be a bit much information?
OTOH, maybe it needs to be mentioned somewhere briefly at least.

seberg (commented):
Some slightly random comments, to take or leave as you like, I only had a fairly brief look.  Basically just thoughts while reading.  I do think mentioning "pseudo" random might be good as a concept to introduce/drop?  It is sometimes a bit too judgy for me, I would rather stress best practices like "only add seeding when you absolutely must freeze for reproducibility", rather than "seeding is bad" :).

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+deterministic anymore. This is what we mean by thread safety.
+
+NumPy came up with a solution: `np.random.RandomState`. The idea is,
+knowing random number generators have a state, let's define them as
+objects. This way, each object can have it's internal state instead of relying
+on a global one. This API is more explicit and prevents any "contamination"
+of state. See this example.
+
+```python
+# the following is wrong and only exists for demonstration purposes!
```
tirthasheshpatel, line 93: Is this still wrong? This seems like the correct way from the description above and below.

jarrodmillman:
@rkern @bashtage Any comments / suggestions?

rkern:
This is not the overall message that I would write for the site's blog, no. The new design was in large part to make seeding easier for _everyone_ to do correctly (at least to write code that _allows_ users to seed), not to tuck it away as a dark art for advanced users. Everyone's entitled to their opinion on these topics, but I don't think this is an expression of the numpy team's opinion on how these features should be used.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+        for future in concurrent.futures.as_completed(futures):
+            sample.append(future.result())
+
+    print(sample[-1])
+
+# [0.33622174]
+# [0.5430262]
+# [0.96671784]
+# [0.37674972]
+# [0.96671784]
```
bashtage, line 80: Mildly curious but you literally get the same value here, which shows the danger of using global in anything other than vanilla single threaded/process code.

content/posts/numpy/generator-rng/index.md
bashtage, line 122: In generally seeding isn't needed or particulary beneficial for existing `RandomState` objects.  Might as well create a new instance with the seed value than calling `seed`.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+here. See [this article]({{< relref "../../scipy/qmc-basics" >}})
+for more information on this.
+
+But why does it matter? How bad this is will depend on your application.
+If you don't, statistical properties are lost. This is hard to get because for
+most people, random is random. So taking an arbitrary draw of sample should not
+matter. And it does not **if** you replicate your analysis using different
+**random** seeds. This is the part everyone is forgetting. Whether you are
+optimizing something or calculating some statistics, you could get (un)lucky
+if you do this just once.
```
bashtage, line 226: Can you include something on the right way to get a high quality (128 or 256 bit) seed at "random", that could then be stored to reproduce later.

content/posts/numpy/generator-rng/index.md
bashtage: Agree.  Nothing in `default_rng` should be considered long-termstable, even the  `BitGenerator`

content/posts/numpy/generator-rng/index.md
bashtage, line 272: I don't agree with this. For example, in academic poblishing you might need to provide code the reproduces statistical methods that rely on Monte Carlo methods (e.g., the Bootstrap). In these cases you also need seed.  The key here IMO is to use a high-quality as-if random 256bit seed, which is like calling `default_rng()` and then storing the value used to seed the first `Generator` and then reusing this when reproducing.

content/posts/numpy/generator-rng/index.md
bashtage: Also `MultiProcess`, which never really works with the global `RandomState`.

bashtage (commented):
Some thoughts.  Always helpful to have more out there about this.

tupui:
Hi everyone, first of all thank you for sharing your thoughts on this draft.

After having read your suggestions and talked to @InessaPawson , I realized that my opinions on seeding might be too controversial for this post. As it is, I understand that it would need a non-endorsement disclaimer from NumPy's maintainers, which is not wanted–for this specific post–as this work was a request I received from NumPy's team.

On that note, I would like to bring to everyone's attention that this is a blog for the community and not the blog of a specific project. To me, it is fine to share opinions even if sometimes there are controversial. It should even be encouraged as it pushes people to encourage, to be involved and also share their opinions. Just like everyone did right now. Hence, maybe we can think about having different kinds of posts and flag them accordingly. We would have opinion posts, endorsed posts, etc. (It could also just be reflected in the keywords we use.) What I want to say is: we want people to contribute posts, and it should be easier to write here than to contribute a tutorial or other documentation in projects.

Back to this post. I propose to rephrase the seeding part by: removing the narrative about the forest and that you should not seed. Doing so, the focus of the article would be solely on using the new generator interface. Still, I would like to add a warning or note about potential risk that you could face doing so.

On the seeding. The main argument I read here and elsewhere about seeding is about reproducibility (in science here). To me, this is a real issue. When you design an experiment and want to publish results, yes you want to be able to replicate them. But not the RNG part, it's the results and conclusions which have to be replicable. And to do this, you have to replicate your study, use bootstrap, confidence intervals, uncertainty propagation, etc.

To me, the bit to bit reproducibility for a (scientific) software is only useful for debugging. I was also using some seeding when I was doing CFD, really helpful to understand why the model didn't do what we expected, or spot difference due to architectures. But when it was show time, I removed the seeds and repeated the simulations several times to have confidence (with confidence intervals) in the output. Look at things like the LHC, they replicate over and over again the same measures to get enough statistics.

As for papers in (Q)MC, I honestly never read anyone fixing a seed. In technometrics and alike, when numbers are reported, you usually see confidence intervals and reviewers would be picky on the number of replicates that was used (e.g. 999 vs 1000, we had this discussion with Art Owen and Fred Hickernell when writing the bootstrap method in SciPy). I really doubt a paper could be accepted in such journals if you report using a given seed to obtain results. Which in my opinion would be the right thing to do. Surprisingly, what you do see in the same journals, is a different rigour around timing/benchmarking. My interpretation on such things is: it's difficult to share best practices among fields (in this case, how to benchmark). Lastly, seeing all the p-value hacking drama is fairly convincing me that using anything like a seed would be seen as a way to control/temper the outcome.

As usual, I am always happy to discuss things with you all–here, in a call or another medium.

content/posts/numpy/generator-rng/index.md
seberg: The last example I can think of that I had, was using a t-SNE (a stochastic embedding) on a smaller dataset.  Of course I have to make sure that the analysis is robust, but I still want to be able to plot an example and be able to replot it without getting a totally different image.

And for larger numerical studies I had intermediate results so large that I did not keep them.  But I still want to be able to "go back" and look at those intermediate results (effectively videos) in case there was something interesting/odd.  So seeding is necessary (although it can be a truly random seed!).
And even months of computer time used, doesn't mean that the final result won't have just few data-points.  So quantitative reproduction would still require seeding (and using the same software version etc.)...

I wonder what the "middle ground" is, since I suspect there is a good reason for the strong opinion:
* You can probably recommend high quality (even truly random) seeds in general.  Which will also prevent cargo-culting a seed into every script, since it requires first generating a seed (i.e. a two step process).
* It may make sense to recommend avoiding to seed until a need for reproducibility is clear?  That way, if you re-run an analysis anyway, you will naturally learn how much the results depends on the random stream.  Arguable, it is mainly a trick to avoid forgetting that even the random stream may be a source of bias(?) that needs to be considered.

InessaPawson:
> I propose to rephrase the seeding part by: removing the narrative about the forest and that you should not seed. Doing so, the focus of the article would be solely on using the new generator interface. Still, I would like to add a warning or note about potential risk that you could face doing so.

@tupui This sounds like a good solution to move forward. Thank you so much for all your work on this article!

content/posts/numpy/generator-rng/index.md
rkern: It's not that much of a surprise. It's getting reseeded to the same state in the main thread at the top of each iteration. If it were seeded only once before the loop and we still got repeats, that _would_ be more of a problem, but the design of even `RandomState` doesn't allow that problem (thanks to the GIL and Cythonized code).

This is something of a contrived example. The main problem with threads and the global in real problems is the loss of determinism, but much of the article is arguing against determinism. If one were committed to seedlessness, there really wouldn't be a problem with using the `np.random` functions and threads even with the hidden global.

content/posts/numpy/generator-rng/index.md
rkern: I think it's too strong and misdirected at the writers of such code. If anyone was wrong, it was us (half me for putting the global in the first place, half someone else who added `np.random.seed()` back in the mists of Subversion). People are just using the APIs we designed for them. It is, indeed, suboptimal code, IMO, but I don't think it's wrong. For many purposes, it's quite fine.

content/posts/numpy/generator-rng/index.md
rkern: There's nothing wrong with the results that are attributable to seeding (or even this one seed). They are just perfectly fine PRNG samples from those distributions. The badness being referred to is just the difference between PRNG samples and the more even, non-independent QRNG samples. But again, unrelated to seeding.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+object based API and not the legacy approach with global state.
+
+Now, one thing with tests is that we want them to be reproducible so that
+we can inspect our code and find issues. Here you want to use a seed.
+This is the **only** valid use for a seed. (Sure if you are an expert in
+Monte Carlo sampling, you can probably find your way... Just don't try this at
+home).
+
+But because you can use a seed doesn't mean you can resort to our old friends
+0, 123456, etc. NumPy proposes a canonical way to randomly give you a seed.
```
rkern, line 277: For these testing applications that you are talking about, any seed will do, really.

rkern:
> To me, it is fine to share opinions even if sometimes there are controversial.

Agreed, with appropriate language when those opinions are more personal and aren't representative of the projects they are about (which this blog is a home for, and as this one was commissioned to be, apparently).

> I propose to rephrase the seeding part by: removing the narrative about the forest and that you should not seed. Doing so, the focus of the article would be solely on using the new generator interface. Still, I would like to add a warning or note about potential risk that you could face doing so.

I don't know what specific risk you are referring to, but I suspect I'll disagree with pretty strongly with that one too (for a project-commissioned article). I don't see much in the `Goodbye seed` section that I think represents the project's viewpoint on the subject, to the extent it has one. For the record, the problem that you introduced with the forest narrative (reusing small-integer and common "joke" seeds) is a reasonable one that I do agree with, but I don't think it was supported by the examples (which digressed to the merits of PRNGs vs QRNGs) or in turn is enough to support the final conclusion of going fully seedless.

The question is, if you delete that section, is there a coherent article left that you actually want to write?

content/posts/numpy/generator-rng/index.md
ricardoprins: I second that here. This can be unclear to some audiences.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+matter. And it does not **if** you replicate your analysis using different
+**random** seeds. This is the part everyone is forgetting. Whether you are
+optimizing something or calculating some statistics, you could get (un)lucky
+if you do this just once.
+
+If your application needs random numbers, then do use random numbers and play
+with the fact that they are random. You do science or anything critical
+using these randomly generated numbers, then please stay away from seeds.
+
+Some will argue that seeding is just used by developers for testing, well it is
```
ricardoprins, line 232: I think this could be rewritten:

"Although it can be argued that seeding is just used by developers for testing, this is not always the case."

ricardoprins: The attempts during this blog to use a more casual language may end up adding unnecessary information that just makes the article longer. I believe that the lines 232-236 can be summarized in a single sentence, for instance.

content/posts/numpy/generator-rng/index.md
```diff
@@ -0,0 +1,304 @@
...
+The worst part here is that people tend to use the same numbers. Some
+copy/pasting might be at play there as well.
+
+Think about it for a while. If everybody is using a seed, which is the same,
+and the original intent was to interact with random numbers for statistical
+analysis or optimization. Do you see the bias that is being introduced?
+
+## Testing with random numbers
+
+Before showing any code though, ask yourself if you actually need ot use
```
ricardoprins, line 244: Lines 244-250 could also be rewritten into a more direct language.

ricardoprins (commented):
I like the content. I think that there should be a little more work on clearing unnecessary comments and information that is conveyed in the text. I also think that it should be adjusted so that it shows the consequences of using seed numbers rather than just emphasizing them as being something bad or incorrect. This is not the case - they do have their usefulness.

stefanv:
> The question is, if you delete that section, is there a coherent article left that you actually want to write?

@tupui Thoughts?

tupui:
Ok all, I removed the long narrative saying seed is bad. I have a more neutral small blob around seeding now. Hope this suits everyone better.

content/posts/numpy/generator-rng/index.md
tupui: I added some notes bellow

content/posts/numpy/generator-rng/index.md
tupui: I added clearer comments

content/posts/numpy/generator-rng/index.md
tupui: Yes, but if people remember only one thing, might be better if they remember the better way.

content/posts/numpy/generator-rng/index.md
tupui: Meaning you don't know what you are actually fixing or not. In the process you might forget some use of `random.random` and then the global seed will have some effect while maybe you think you are playing with an object.

rkern (commented):
I'm sorry, but the fundamental issues I raised are not fixed. This article is still a polemic against PRNGs, not "Random numbers done right". It does not support its cases with coherent arguments, much less ones that I think are strong. It does not give the user much practical advice on using the PRNG APIs well, mostly just warning people away from them. Softening the language in a few places did not change this fundamental problem. It seems more like those changes were done to appease reviewers, not because the new language is part of a coherent story being told that the author believes in. The article is no more coherent because of them.

To the extent that I have any influence on the matter as one of the primary architects of `numpy.random`, I don't think this article can go out without a non-endorsement disclaimer. I don't think it represents the views of anyone who worked on its design. I would rather it not go out on `scientific-python.org` at all.

FWIW, I think @albertcthomas has a [fine article published on his own](https://albertcthomas.github.io/good-practices-random-number-generators/) that pretty much represents the project's view on matters (with some nitpicks, a section on not using short dummy seeds and how to get nice long ones from system entropy would be good improvement, but all told, that's a fairly minor concern).

content/posts/numpy/generator-rng/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Random numbers done right"
-date: 2022-04-25
+date: 2022-12-28
 draft: false
 description: "
 Use `np.random.Generator` and don't seed! A quick explainer on random number
```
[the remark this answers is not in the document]
tupui: It's a leftover.

content/posts/numpy/generator-rng/index.md
```diff
@@ -160,20 +162,11 @@ For more details have a look at the documentation of
...
-robust to the environment.
-
-Setting the seed of a random number generator is just like that. There is
-nothing random anymore, you are breaking your random number generator, breaking
-your forest. You see often in **code** the exact same seeds: 0, 123456.
-In fact, if you do this, you are using a fixed sequence of points which
-effectively has no guarantee over the quality of the points.
+A quick note about using a seed. When you use a seed, you are using a fixed
+sequence of points which effectively has no guarantee over the quality of the
+points.
```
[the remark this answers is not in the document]
tupui: > The quality of the sequence of points actually drawn in one invocation of the program is mostly down to the PRNG algorithm

Which is the point. And I go to say that if you do want a single sample, then use QMC which would give you a better sequence. If not and your program does produce different draw without a seed, then you will have MC on average with its convergence rate. Then ok (if you are fine with this rate OC).

content/posts/numpy/generator-rng/index.md
```diff
@@ -269,7 +249,7 @@ object based API and not the legacy approach with global state.
 
 Now, one thing with tests is that we want them to be reproducible so that
 we can inspect our code and find issues. Here you want to use a seed.
-This is the **only** valid use for a seed. (Sure if you are an expert in
+This is one of the few valid use for a seed. (Sure if you are an expert in
```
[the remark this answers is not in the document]
tupui: I don't know. Personally I think none. This was some softening and the rest can be removed.

Others or you can suggest other applications.

tupui:
> I'm sorry, but the fundamental issues I raised are not fixed. This article is still a polemic against PRNGs, not "Random numbers done right". It does not support its cases with coherent arguments, much less ones that I think are strong. It does not give the user much practical advice on using the PRNG APIs well, mostly just warning people away from them. Softening the language in a few places did not change this fundamental problem. It seems more like those changes were done to appease reviewers, not because the new language is part of a coherent story being told that the author believes in. The article is no more coherent because of them.

I frankly disagree. Of course I updated the article based mostly on your criticism. But I am not reading your mind as well... If I read the article in its current version, I am only saying that if you use a seed, it will give you the same sequence of point which might not be great. It's not saying anymore that I think it's a crapy idea and should never be done in any context especially if you are doing science.

The example that you don't like, the one with the distributions, is just trying to say: look you fixed the seed, this is giving you these distributions which are not optimal. Yes OC they are samples from the correct distributions, I never say they weren't, I am saying they are poor samples from the distributions. Hence if you don't replicate your experiment and just use a single sample you have bad stats.

And I conclude the section saying that one need to be aware of that and propose to use QMC for people that really want a fixed sequence.

That might be rephrased, sure. But I disagree when you are saying that the example is wrong and cannot be used to show what I had intended for it.

> To the extent that I have any influence on the matter as one of the primary architects of `numpy.random`, I don't think this article can go out without a non-endorsement disclaimer. I don't think it represents the views of anyone who worked on its design. I would rather it not go out on `scientific-python.org` at all.

We never said that any article posted on this blog would be endorsed by any projects. It's also why this has my own name and not "NumPy's team". We can maybe add a general banner on the blog about that.

To be clear, in this case if I were to add some non-endorsement banners, well then I would prefer to go back to my opinionated text because now it has my name and I don't endorse saying a seed is ok to use for anything else than testing.

> FWIW, I think @albertcthomas has a [fine article published on his own](https://albertcthomas.github.io/good-practices-random-number-generators/) that pretty much represents the project's view on matters (with some nitpicks, a section on not using short dummy seeds and how to get nice long ones from system entropy would be good improvement, but all told, that's a fairly minor concern).

Following this exchange, I see 2 options moving forward:

1. You do edit suggestions (real sentences, not something I have to interpret wrongly as you suggest I have been doing up until now)
2. I close the PR.

rkern:
You said that this was commissioned by the NumPy team and that a non-endorsement disclaimer would be undesirable because of that. I was taking you at your word on that. I don't know any of the details of that commission, but I doubt anyone asked for a polemic against seeding PRNGs or avoiding PRNGs. I don't have any editorial control over the blog, but I would suggest to those who do to encourage you to publish your article elsewhere, not just under your own name here. I do not think it is quality content; it's not just that I disagree with its points.

I won't be making sentence-level rewrites; as I said, I think the whole article is incoherent. I made comments about individual items not as suggestions for individual fixes but arguments against publishing the article at all. If the NumPy project wants to commission an article on "Random numbers done right", giving practical advice on the usage of the new PRNG API, I would just ask @albertcthomas to republish his. I think it's really good.

PR #76, opened 2022-06-15
Add pre-commit and Hugo checks to makefile

tupui:


Makefile
```diff
@@ -9,6 +9,8 @@ help:   ## show this help
 
 prepare:
 	git submodule update --init
+	pre-commit install || echo "Install 'pre-commit' in a Python environment with: 'pip install pre-commit'"
+	hugo version. || echo "Install Hugo from: https://gohugo.io"
```
tupui, line 13:
```suggestion
	hugo version || echo "Install Hugo from: https://gohugo.io"
```

jarrodmillman (approved):

PR #80, opened 2022-07-06
Updates on VF2++

kpetridis24:
This is the second post from my GSoC 2022 journey. It includes updates of my progress so far. More updates in the next posts.


content/posts/networkx/node-ordering-Ti-updating/index.md
```diff
@@ -0,0 +1,91 @@
...
+description: "
+Summary of the progress on VF2++
+"
+tags: ["gsoc", "networkx"]
+displayInList: true
+author: ["Konstantinos Petridis"]
+
+---
+
+This post includes all the major updates since the last post about VF2++. Each section is dedicated to a different
```
tupui, line 14: Add a link to the previous post. You can have relative links `[here]({{< relref "../GSoC-2022" >}}`.

I would also update the first post with a table of content at the beginning. We can have the same table in all posts. Something similar is found in other blogs when there are follow up.

content/posts/networkx/node-ordering-Ti-updating/index.md
```diff
@@ -0,0 +1,91 @@
...
+where $V_1$ and $E_1$ contain all the nodes and edges of the first graph respectively, and $m$ is a dictionary, mapping
+every node of the first graph to a node of the second graph. Now if we interpet the above equation, we conclude that
+$T_1$ contains uncovered neighbors of covered nodes. In simple terms, it includes all the nodes that do not belong in
+the mapping $m$ yet, but are neighbors of nodes that are in the mapping. In addition,
+
+$$\tilde{T_1}=\(V_1 \setminus m \setminus T_1\)$$
+
+The following figure is meant to provide some visual explanation of what exactly $T_i$ is.
+
+<center><img src="Ti.png" alt="Illustration of $T_i$"/></center>
```
tupui, line 72: Is there a reason you did not use normal markdown here?
```suggestion
![Illustration of $T_i$.](Ti.png)
```

tupui (commented):
Thank you for the update @kpetridis24.

As @rossbar suggested previously, we should probably group all these posts under a `gsoc` folder (either in `posts/` or `posts/networkx`).

Otherwise it reads well. Not sure how much detail we want to give. I left some suggestions.

content/posts/networkx/node-ordering-Ti-updating/index.md
kpetridis24: nope, i just took it as is from another post. I am not an md expert unfortunately :)

kpetridis24:
> Thank you for the update @kpetridis24.
> 
> As @rossbar suggested previously, we should probably group all these posts under a `gsoc` folder (either in `posts/` or `posts/networkx`).
> 
> Otherwise it reads well. Not sure how much detail we want to give. I left some suggestions.

I could provide a lot more detail but i didn't consider it necessary. I took a look at previous posts and they also kind of include details on the project and code from the implementations. So i thought that some detailed information would probably fit.

kpetridis24:
@tupui @rossbar I added a diagram showing the performance improvement when using the incremental updating and i also included some complexity analysis. I tried to keep the complexity section VF2++-agnostic, particularly i just presented the two functions and computed the bigO for both of them, something the reader can do as well without any previous knowledge on the implementation. Feel free to give any feedback.

content/posts/networkx/node-ordering-Ti-updating/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+    return T1, T2, T1_out, T2_out
+```
+
+which based on the previous notation, is:
+
+$$O(2D + 2(D + M_{T_1}) + 2D) = O(D + M_{T_1})$$
+
+where $M_{T_1}$ is the expected (average) number of elements in $T_1$. Certainly, the complexity is much better in this
+case, as $D$ and $M_{T_1}$ are significantly smaller than $N_mD$ and $N$.
```
tupui, line 155: The conclusion should stand out
```suggestion
where $M_{T_1}$ is the expected (average) number of elements in $T_1$.

Certainly, the complexity is much better in this
case, as $D$ and $M_{T_1}$ are significantly smaller than $N_mD$ and $N$.
```

content/posts/networkx/node-ordering-Ti-updating/index.md
```diff
@@ -0,0 +1,91 @@
...
+observations to implement an incremental computation of $T_i$ and $\tilde{T_i}$ and make VF2++ more efficient.
+
+- $T_i$ is empty in the beggining, since there are no mapped nodes ($m=\varnothing$) and therefore no neighbors of
+  mapped nodes.
+- $\tilde{T_i}$ initially contains all the nodes from graph $G_i, i=1,2$ which can be realized directly from the
+  notation if we consider both $m$ and $T_1$ empty sets.
+- Every step of the algorithm either adds one node $u$ to the mapping or pops one from it.
+
+We can conclude that in every step, $T_i$ and $\tilde{T_i}$ can be incrementally updated. This method avoids a ton of
+redundant operations and results in significant performance improvement.
```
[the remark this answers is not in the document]
tupui: Thank you for the update. If you can regenerate the figure, we need to make it more accessible by using different linestyles and adding a legend.

kpetridis24:
@tupui Added different linestyles and legend!

Makefile
```diff
@@ -36,7 +36,7 @@ html: prepare
 
 serve: ## Serve site, typically on http://localhost:1313
 serve: prepare
-	@hugo --printI18nWarnings server
+	@hugo server
```
tupui, line 39: Revert this too.
```suggestion
	@hugo --printI18nWarnings server
```

tupui (commented):
LGTM, I will let @rossbar or another NetworkX maintainer have a look before merging.

content/posts/networkx/node-ordering-Ti-updating/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+![Performance comparison between brute force Ti computing and incremental updating.](acceleration.png)
+
+The above graph shows the difference in performance between using the exhaustive brute force and incrementally updating
+$T_i$ and $\tilde{T_i}$. The graph used to obtain these measurements was a regular
+[GNP Graph](https://en.wikipedia.org/wiki/Erd%C5%91s%E2%80%93R%C3%A9nyi_model) with a probability for an edge equal to
+$0.7$. It is directly observed that the increase in time when using the brute force method, follows an exponential rate,
+while in our method, it's not significantly affected by incrementing the number of nodes. Let's talk time complexity.
```
rossbar, line 99: I think this statement doesn't *quite* match what the performance plot shows. For instance, both methods *appear* to have an exponential increase in run-time vs. number of nodes, it's just that the incremental update method increases much less rapidly. Perhaps something like:

```suggestion
$0.7$. It can clearly be seen that execution time of the brute force
method increases much more rapidly with the number of nodes/edges than
the incremental update method, as expected.
```

content/posts/networkx/node-ordering-Ti-updating/index.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+which based on the previous notation, is:
+
+$$O(2D + 2(D + M_{T_1}) + 2D) = O(D + M_{T_1})$$
+
+where $M_{T_1}$ is the expected (average) number of elements in $T_1$.
+
+Certainly, the complexity is much better in this
+case, as $D$ and $M_{T_1}$ are significantly smaller than $N_mD$ and $N$.
```
rossbar, line 157: Since you're already working on the next post, it might be nice to wrap this one up in a sentence or two and give readers a teaser about what's coming next! e.g. `In this post we've investigated how node ordering is implemented for the VF2++ algorithm; next we'll investigate another important compenent: cuts and consistency checks <or whatever your next post will be about)`

rossbar (commented):
This is great, thanks @kpetridis24 !

I've left just two final suggestions (and opened a PR against your fork with some minor grammatical/formatting changes). Other than that I think this is good to go - great job summarizing your work on this component of the VF2++ algorithm!

content/posts/networkx/node-ordering-Ti-updating/index.md
kpetridis24: Sure, you are right!

rossbar (approved):
> The formatting destroyed the pseudocodes :')

Yeah, I don't know what happened in 1f86fd3... whatever it was it was a nasty change.

Anyways, I went ahead and reverted that mess, cherry-picked out the conclusion that you added and re-committed it (w/ a grammatical edit or two) in 5486851. So the formatting errors should be fixed and all of your content present and accounted for!

Thanks @kpetridis24 for writing it up!

PR #83, opened 2022-07-11
Reorganize NetworkX blog posts

rossbar:
All of the NetworkX-specific posts that we have on the blog so far have come from GSoC contributors (thanks @mjschwenne and @kpetridis24 !). Since these posts are all related to a specific project or algorithm, I've re-organized the `posts/networkx` directory so that all of the posts live in a project-specific directory, e.g. `aTSP` for Matt's project and `vf2pp` for Konstantinos' project.

I've also added a README in an effort to explain this organization for future GSoC/related students who are using the blog to host their project progress.

Finally, I've added project-specific tags to the posts to make it easier to find all of the posts related to a given GSoC project.

PR #85, opened 2022-07-12
NetworkX GSoC 2022 Blog Post 3: VF2++ Candidates and Feasibility

kpetridis24:
This post discusses the implementation of the final helpers of the VF2++ algorithm, specifically the feasibility checks and the candidate pair selection.


Makefile
```diff
@@ -36,7 +36,7 @@ html: prepare
 
 serve: ## Serve site, typically on http://localhost:1313
 serve: prepare
-	@hugo --printI18nWarnings server
+	@hugo server
```
tupui, line 39: Please revert this.

content/posts/networkx/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+**we compute the candidates**
+
+$$v\in V_2$$
+
+where $V_1$ and $V_2$ are the nodes of $G_1$ and $G_2$ respectively. Now this is a puzzle that does not require a lot of
+specific knowledge on graphs or the algorithm itself. Keep up with me, and you will realize it yourself. First, let $M$
+be the mapping so far, which includes all the "covered nodes" until this point. There are actually **three** different
+types of $u$ nodes that we might encounter.
+
+# Case 1
```
tupui, line 53: I believe you want to increase the indentation level to be 3. So `###`

content/posts/networkx/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+isomorphism.
+
+## ISO Feasibility Rules
+
+Let's assume that given a node $u$, we obtained its candidate $v$ following the process described in the previous section.
+At this point, the **Feasibility Rules** are going to determine whether the mapping should be extended by the pair $u-v$
+or if we should try another candidate. The **feasibility** of a pair $u-v$ is examined by **consistency** and
+**cutting** checks.
+
+# Consistency rules
```
tupui, line 89: Same here I think it should be 3.

tupui (changes requested):
Thank you for the update @kpetridis24!

I have a few suggestions. Please remove the update to Ti.png as the size increased and I don't see any difference. Also, #83 will go in first hence can you follow the new directory tree?

rossbar:
Just a heads up @kpetridis24 , I changed the underlying organization of the networkx posts by adding an extra layer to the filesystem hierarchy for project-specific posts. Unfortunately this means I created a merge conflict for you :)

The good news is it should be relatively easy to resolve - you can merge with `main`, then delete the `node-ordering-Ti-updating/Ti.png` image to resolve the conflict. Once the merge succeeds, you can add the image back at the new location in `vf2pp/node-ordering-Ti-updating/Ti.png`.

After that, you should move the whole `ISO-feasibility-candidates` directory into the new `vf2pp` project directory, i.e. `git mv ISO-feasibility-candidates vf2pp`.

At that point, you should be all set and can refocus on the post itself. LMK if you have any trouble with the above procedure - I'd be happy to take care of the conflict locally and push to your branch if you get stuck.

kpetridis24:
@tupui @rossbar I think i fixed everything!

content/posts/networkx/vf2pp/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+from the first graph is matched to a node from the second graph. This searching for "feasible pairs" can be visualized
+by a tree, where each node is the candidate pair that we should examine. This can become much clearer if we take a look
+at the below figure.
+
+<center><img src="dfs.png" alt="DFS VF2++ example$"/></center>
+
+In order to check if the graphs $G_1$, $G_2$ are isomorphic, we check every candidate pair of nodes and if it is
+feasible, we extend the mapping and go deeper into the tree of pairs. If it's not feasible, we climb up and follow a
+different branch, until the mapping is extended to its full. In our example, we start by examining node 0 from G1, with
+node 0 of G2. After some checks that are skipped for simplicity (but can be perceived by observation), we decide that the
```
rossbar, line 31: For the sake of brevity, we could instead implore the reader to keep reading for details!

```suggestion
node 0 of G2. After some checks (details below), we decide that the
```

content/posts/networkx/vf2pp/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+nodes 0 and 0 are matching, so we go deeper to map the remaining nodes. The next pair is 1-1, which fails the
```
rossbar, line 32: This part is actually a bit confusing because there is no node `1` in G_2. Of course - this is the whole point: it's not feasible to map node `1` in G_1 to node `1` in G_2 because node `1` in G_2 doesn't exist.

In fact, is this a good example? AIUI from reading on, node `1` in G_2 actually *wouldn't* be part of the candidate nodes in G_2 because the first prerequisite for creating the candidate set is that the nodes must exist in G_2, right?

Perhaps the example could be updated to use `1-3` as the next candidate, which would not be feasible because of the degree mismatch.

Unfortunately I think this may also necessitate updating the left-most figure in the accompanying image, as I don't think checking `1-1` makes sense --- unless I'm completely misunderstanding candidate selection!

content/posts/networkx/vf2pp/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+$$v\in V_2$$
+
+where $V_1$ and $V_2$ are the nodes of $G_1$ and $G_2$ respectively. Now this is a puzzle that does not require a lot of
+specific knowledge on graphs or the algorithm itself. Keep up with me, and you will realize it yourself. First, let $M$
+be the mapping so far, which includes all the "covered nodes" until this point. There are actually **three** different
+types of $u$ nodes that we might encounter.
+
+### Case 1
+
+Node $u$ has no neighbors (degree of $u$ equals to zero). It can directly be realized that it would be redundant to test
```
rossbar, line 55: More (unimportant) wording suggestions, feel free to ignore!
```suggestion
Node $u$ has no neighbors (degree of $u$ equals to zero). It would be redundant to test
```

content/posts/networkx/vf2pp/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+as candidates for $u$, nodes from $G_2$ that have more than zero neighbors. That said, we eliminate most of the possible
+candidates and keep those that have the same degree as $u$ (in this case, zero). Pretty easy right?
+
+### Case 2
+
+Node $u$ has neighbors, but none of them belong to the mapping. This situation is illustrated in the following figure.
+
+<center><img src="c2.png" alt="candidates"/></center>
+
+The grey lines indicate that the nodes of G1 (left 1,2) are mapped to the nodes of G2 (right 1,2). They are basically
```
rossbar, line 65: Just for the sake of consistency
```suggestion
The grey lines indicate that the nodes of $G_1$ (left 1,2) are mapped to the nodes of $G_2$ (right 1,2). They are basically
```

content/posts/networkx/vf2pp/ISO-feasibility-candidates/index.md
```diff
@@ -0,0 +1,156 @@
...
+### Case 3
+
+Node $u$ has neighbors and some of them belong to the mapping. This scenario is also depicted in the below figure.
+
+<center><img src="c3.png" alt="candidates"/></center>
+
+In this case, to obtain the candidates for $u$, we must look into the neighborhoods of nodes from $G_2$, which map back
+to the covered neighbors of $u$. In our example, $u$ has one covered neighbor (1), and 1 from G1 maps to 1 from G2,
+which has $v$ as neighbor. Also, for v to be considered as candidate, it should have the same degree as $u$, obviously.
+Notice how every node that is not in the neighborhood of 1 (in G2) cannot be matched to $u$ without breaking the
```
rossbar, line 79: More math formatting consistency nits...
```suggestion
In this case, to obtain the candidates for $u$, we must look into the neighborhoods of nodes from $G_2$, which map back
to the covered neighbors of $u$. In our example, $u$ has one covered neighbor (1), and 1 from $G_1$ maps to 1 from $G_2$,
which has $v$ as neighbor. Also, for v to be considered as candidate, it should have the same degree as $u$, obviously.
Notice how every node that is not in the neighborhood of 1 (in $G_2$) cannot be matched to $u$ without breaking the
```

rossbar (commented):
Thanks @kpetridis24 , this is a very nice writeup!

I took the liberty of making a few small wording/formatting/grammatical changes, so make sure to pull before your next push!

I've also left a few wording suggestions, most of which are just suggestions and can be safely ignored.

There is *one* thing though that I think needs to be cleared up related to the first illustration which is used to explain the candidate selection. The illustration (and surrounding text) states that the "next" candidate pair one might check after `0-0` is `1-1`. This makes sense at face value, given that `1` is the next neighbor of `0` in `G_1`; however, I don't think this makes sense in the context of the *actual* node candidate selection procedure. The main sticking point is that candidate selection is limited *only* to those nodes that exist in `G_2`. Since G_2 doesn't have a node `1`, `1-1` would never be selected as a candidate node pair, right?

I think your approach of using simple visual examples to explain the procedures of the algorithm is incredibly useful, but we should be careful that the examples match what is actually performed in the algorithm itself!

OTOH, I could just be understanding things incorrectly... if so please let me know!

content/posts/networkx/vf2pp/ISO-feasibility-candidates/index.md
kpetridis24: No you are right, i was too lazy to check. Your understanding is fine. Basically we would try to match nodes 1-1 if 1 was present in G2 (which i didn't notice it wasn't) so i will just relabel the nodes in G2 or just pick another candidate for 1. I wanted to shift the emphasis on the logic of the procedure rather than the details, but this could indeed be misleading. I will correct when immediately. Thanks for noticing!

kpetridis24: All done!

rossbar (approved):
Thanks for your patience & nice explanations @kpetridis24 , this is a very nice post - it certainly contributes to my understanding of what's going on in VF2++!

PR #92, opened 2022-08-25
VF2++

kpetridis24:
This is the 4th post, discussing the implementation of VF2++ and presenting some benchmarks.


content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,132 @@
...
+
+---
+
+The last and final post discussing the **VF2++ helpers** can be found [here]({{< relref "../ISO-feasibility-candidates" >}}).
+Now that we've figured out how to solve all the sub-problems that **VF2++** consists of, we are ready to combine our
+implemented functionalities to create the final solver, for the **Graph Isomorphism** problem.
+
+## Introduction
+
+We should quickly remind at this point, that the individual functionalities already implemented are:
```
rossbar, line 20: Just a wording suggestion, feel free to ignore!

```suggestion
We should quickly review the individual functionalities used in the VF2++ algorithm:
```

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,132 @@
...
+>    - $n,nc=stack.peek()$
+>    - **if** $nc$ is empty **do**
+>      - $stack.pop()$
+>      - $\def\textsc#1{\dosc#1\csod}{restoreState()}$
+>      - $continue$
+>    - **if** $\def\textsc#1{\dosc#1\csod}{feasibility(n, nc.pop())}$ **do**
+>      - **if** len(m) equals len(G1) **return True**
+>      - $\def\textsc#1{\dosc#1\csod}{updateState()}$
+>      - $nextNode=order.next()$
+>      - $stack.append\(\(nextNode, \def\textsc#1{\dosc#1\csod}{findCandidates(nextNode)}\)\)$
```
rossbar, line 66: One of the great things about Python is it's readability - in principle it'd be great if, instead of "rewriting" the algorithm in pseudocode, we could instead show the actual implementation!

For example, [the actual implementation](https://github.com/networkx/networkx/blob/bffcd74649fb95a57fb834846eb3c7d9693c55b8/networkx/algorithms/isomorphism/vf2pp.py#L171-L236) is only slightly longer, and IMO is quite readable. I'm not sure whether there's a nice way to embed code directly from github, or whether you'd have to copy it into a code-block, but I think that'd be a nice improvement.

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,132 @@
...
+**random graphs** without labels, for number of nodes anywhere between the range $\(100-2000\)$. The results are depicted
+in the two following diagrams.
+
+<center><img src="times.png" alt="vf2++ and vf2 times"/></center>
+<center><img src="speedup.png" alt="speedup"/></center>
+
+We notice that the maximum speedup achieved is **14x**, and it keeps increasing gradually, as the number of nodes increase.
+It is also highly prominent that the increase in number of nodes, doesn't seem to affect the performance of **VF2++** to
+a significant extent, when compared to the drastic impact on the performance of **VF2**. Our results are almost identical
+to those presented in the original **VF2++ paper**, verifying the theoretical analysis and premises of the literature.
```
rossbar, line 80: It might be worth turning the `**VF2++ paper**` into a link to the paper.

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Optimizations
+
+The achieved boost is due to some key improvements and optimizations, and more specifically:
```
rossbar, line 84:
```suggestion
The achieved boost is due to some key improvements and optimizations, specifically:
```

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,132 @@
...
+
+- **Optimal node ordering**, which avoids following unfruitful branches that will result in infeasible states. We make sure that the nodes that have the biggest possibility to match, are accessed first.
+- **Implementation in a non-recursive manner**, which saves both time and space.
```
rossbar, line 87: Recursive implementations are not *always* generally slower than iterative ones, so I'd vote to be a little more specific here. IMO the main advantages of an iterative implementation for this case are:
 - Avoiding Python's maximum recursion depth limit
 - Reducing function call overhead

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,132 @@
...
+    for n in G2_nodes_of_degree[G1.degree[u]].intersection(
+        G2_nodes_of_label[G1_labels[u]]
+    )
+]
+for candidate in candidates:
+    if feasibility(u, candidate):
+        do_stuff()
+```
+
+Immediately we have drastically reduced the number of checks performed and calls to the function, as now we only apply them to nodes of the same degree and label as $u$. This is a simplification for demonstration purposes. In the actual implementation there are more checks and extra shrinking of the candidate set.
```
rossbar, line 132: This post does a really nice job summarizing the algorithm; IMO, what would really knock this out of the park would be to finish with an example of applying VF2++ to a "classic" isomorphism problem.

It's not *necessary*, but I think this would really be a compelling way to end the series of posts. Can we think of good "canonical" example of an isomorphism problem that we could demo here?

kpetridis24: @rossbar 
Do you mean providing an example and following the steps of the algorithm one by one, explaining everything? Yup, we could use the graphs included in the isomorphism wikipedia page!

rossbar: I was thinking even lighter-weight than that - no reason to do the step-by-step breakdown, just show the result of vf2pp in action!

kpetridis24: Ok so for example we can show the mapping returned be the solver or smth

kpetridis24:
@rossbar 
I applied all changes. You could take a look at the demo section i added, to see if it's close to what you were thinking of adding.

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,247 @@
...
+    (6, 7),
+    (7, 8),
+]
+
+# map the nodes between the two graphs
+mapped_nodes = {1: "a", 2: "h", 3: "d", 4: "i", 5: "g", 6: "b", 7: "j", 8: "c"}
+
+# create the two networkx graphs
+G1 = nx.Graph(edges)
+G2 = nx.relabel_nodes(G1, mapped_nodes)
```
rossbar, line 213: From a pedagogical perspective, creating the second graph via a mapping from the first sort of seems like cheating, as then it is immediately obvious the graphs are isomorphic! It's almost like you're giving the answer away at the very beginning of the example :)

Even though it's more tedious, I think this could be improved by constructing both graphs manually. This doubles as a nice opportunity to show how graphs can be constructed from edge lists in networkx:

````suggestion
Let's start by constructing the graphs from the image above. We'll call
the graph on the left `G` and the graph on the left `H`:

```python
import networkx as nx

G = nx.Graph(
    [
        ("a", "g"),
        ("a", "h"),
        ("a", "i"),
        ("g", "b"),
        ("g", "c"),
        ("b", "h"),
        ("b", "j"),
        ("h", "d"),
        ("c", "i"),
        ("c", "j"),
        ("i", "d"),
        ("d", "j"),
    ]
)

H = nx.Graph(
    [
        (1, 2),
        (1, 5),
        (1, 4),
        (2, 6),
        (2, 3),
        (3, 7),
        (3, 4),
        (4, 8),
        (5, 6),
        (5, 8),
        (6, 7),
        (7, 8),
    ]
)
```
````

File: content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md [hunk 8]
```diff
@@ -0,0 +1,247 @@
...
+
+# use the VF2++ without taking labels into consideration
+res = nx.vf2pp_is_isomorphic(G1, G2, node_label=None)
+# res: True
+res = nx.vf2pp_isomorphism(G1, G2, node_label=None)
+# res: {1: "a", 2: "h", 3: "d", 4: "i", 5: "g", 6: "b", 7: "j", 8: "c"}
+```
```
rossbar, line 220: For this section, I think it's worth highlighting all three of the "main" vf2pp functions. There's a nice, natural progression to how they can be presented: start with `vf2pp_is_isomorphic` as a simple binary check for isomorphism, then introduce `vf2pp_isomorphism` to give a single, valid isomorphic mapping. Finally, you can highlight that there may be more than one isomorphic mapping, and users can find *all* isomorphisms with `list(vf2pp_all_isomorphisms)`, noting that it's a generator because there may be *a lot* of isomorphisms!

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,247 @@
...
+G1 = nx.Graph(edges)
+G2 = nx.relabel_nodes(G1, mapped_nodes)
+
+# assign labels to the graphs (same label to mapped nodes)
+nx.set_node_attributes(G1, dict(zip(G1, itertools.cycle(labels))), "label")
+nx.set_node_attributes(
+    G2,
+    dict(zip([mapped_nodes[n] for n in G1.nodes()], itertools.cycle(labels))),
+    "label",
+)
```
rossbar, line 238: This is excellent - I really like the way things are presented first unlabelled then with labels. I have the same general comment as above though re: how the labels are applied: it's sort of "giving away the answer". Again, I'd suggest doing the more tedious manual labelling:

```suggestion
Now if we wanted to take labels into consideration as well, we could do something like:

```python
G_nattrs = {
    "a": "blue",
    "g": "green",
    "b": "pink",
    "h": "red",
    "c": "yellow",
    "i": "orange",
    "d": "cyan",
    "j": "purple",
}
nx.set_node_attributes(G, G_nattrs, name="color")

H_nattrs = {
    1: "blue",
    2: "red",
    3: "cyan",
    4: "orange",
    5: "green",
    6: "pink",
    7: "purple",
    8: "yellow",
}
nx.set_node_attributes(H, H_nattrs, name="color")
```

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
```diff
@@ -0,0 +1,247 @@
...
+
+res = nx.vf2pp_is_isomorphic(G1, G2, node_label="label")
+# res: True
+res = nx.vf2pp_isomorphism(G1, G2, node_label="label")
+# res: {1: "a", 2: "h", 3: "d", 4: "i", 5: "g", 6: "b", 7: "j", 8: "c"}
+```
+
+Notice how in the first case, our solver may return a different mapping every time, since the absence of labels results in nodes that can map to more than one others. For example, node 1 can map to both a and h, since the graph is symmetrical.
+On the second case though, the existence of a single, unique label per node imposes that there's only one match for each node, so the mapping returned is deterministic.
```
rossbar, line 247: Finally, you can make the point about potentially different mappings explicitly here by showing that `nx.vf2pp_all_isomorphisms` yields only a single valid mapping when labels are taken into account, as opposed to the many that are returned in the unlabeled case.

rossbar (commented):
I went ahead and removed the explicit pngs that were added in favor of just directly linking to the images hosted on wikipedia - that will save a bit of space in the repo :). Similarly we should add some alt-text for the two added images!

The example generally looks good, I just had a few suggestions on how it might be framed more for a reader who perhaps didn't already know that the graphs were isomorphic.

content/posts/networkx/vf2pp/graph-iso-vf2pp/index.md
kpetridis24: @rossbar you can take a look when available and see if it's all good.

kpetridis24:
@rossbar @tupui 
So today GSoC officially ends. Is it possible to take a look at this and merge it if its ready so i can provide link to it in my final submission?

tupui:
Hi @kpetridis24, congrats on completing the program 👏 This is in good shape, I am happy to merge once @rossbar approves (not sure you have rights here Ross?).

kpetridis24:
> Hi @kpetridis24, congrats on completing the program clap This is in good shape, I am happy to merge once @rossbar approves (not sure you have rights here Ross? EDIT: actually you should since you are listed as an Editor, so I will let you handle that smiley).

Thanks a lot!!

kpetridis24:
@rossbar two hours for the submission to close :)

rossbar:
> @rossbar two hours for the submission to close :)

No worries @kpetridis24 :) For the record though, whether or not things are merged doesn't matter for GSoC - you can include the PR in the submission!

Issue #118, opened 2023-07-12
spell checker


jarrodmillman:
It might be nice to add a spell checker to the CI/lint checks.

stefanv:
See https://learn.scientific-python.org/development/guides/style/#spelling

PR #123, opened 2023-07-17
Add codespell to github actions and fix typos

MridulS:
closes https://github.com/scientific-python/blog.scientific-python.org/issues/118


tupui (approved):
Noice!

tupui (commented):
ah right, latex and all. Seems like there is also `weigth`

jarrodmillman (approved):

PR #124, opened 2023-07-19
Remove announcements of private attribute `_is_array`

dschult:
Remove the blog comments about using private attribute `_is_array`.


stefanv:
Thanks!

PR #48, opened 2022-03-30
Update frontpage

jarrodmillman:
# Before
screenshot
# After
screenshot

PR #128, opened 2023-07-20
Fix mainpage

jarrodmillman:
This undoes https://github.com/scientific-python/blog.scientific-python.org/pull/48 which I did when I couldn't figure out how to do this. Depends on https://github.com/scientific-python/scientific-python-hugo-theme/pull/249.

This removes more custom code and relies more on the theme. There was an issue with a recent update to the theme that would have caused more customization here, which is why I decided to go back and remove the previous work around. I also never liked all the whitespace it put at the top of the site. It is now easier to see the content!

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -70,12 +70,3 @@ img.post-featuredImage {
   margin-top: 60px;
   padding-bottom: 10px;
 }
-
-.hero-title-content {
-  display: flex;
-  flex-direction: row;
-  width: 100%;
-  max-width: 1200px;
-  margin: 2vh auto;
-  justify-content: space-around;
-}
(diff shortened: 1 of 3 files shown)


brianhawthorne (approved):
LGTM

jarrodmillman:
# Now
2023-07-20T16:41:02,418235458-07:00

 # Before
2023-07-20T16:42:01,965213859-07:00

jarrodmillman:
Once this goes in, I will update learn.scientific-python.org, so all the sites are back in sync with scientific-python.org.

jarrodmillman:
I am merging this now so we can have all three sites synced. Once this is merged, we can start discussing how to improve the look and functionality.

PR #131, opened 2023-08-26
Rename URL of development guide

stefanv:


jarrodmillman:
To be consistent with
- https://blog.scientific-python.org/scientific-python/dev-summit-1/
- https://blog.scientific-python.org/scientific-python/dev-summit-1-sparse/

I would call this
- https://blog.scientific-python.org/scientific-python/dev-summit-1-development-guide/

henryiii:
I'd be fine with either https://blog.scientific-python.org/scientific-python/dev-summit-1-development-guide/ or https://blog.scientific-python.org/scientific-python/dev-summit-1/development-guide. The former does fit better with what's already there, and we don't want to break old URLs, so if it is renamed, you'd also want a redirect. If we do this one fairly quickly, then I think it's fine without a redirect. (I do want to start sharing this link soon!)

stefanv:
I went with https://blog.scientific-python.org/scientific-python/dev-summit-1-development-guide/; currently, we have `project/slug`, so nesting it deeper would require further changes to the configuration.

jarrodmillman (approved):

jarrodmillman:
@henryiii Did you link to or share the original URL? Should we add a redirect? We just need to add an entry to the netlify conf like https://github.com/scipy/scipy.org/blob/main/netlify.toml

PR #135, opened 2023-09-20
Numpy RNGs blog post

albertcthomas:
Blog post on good practices with numpy RNGs (and RNGs in general but focused on numpy features). This is a blog post that I published on my personal page and @stefanv suggested to re-publish it here.

I should make sure it is up-to-date with the recent numpy developments. Some parts could also be changed. I'm opening the PR so we can start discussing about this post.

@stefanv (or anyone else,) let me know if you wanted to add something to this post


content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,110 @@
...
+
+In short:
+* Instead of using `np.random.seed`, which reseeds the already created global numpy RNG and then using `np.random.*` functions you should create a new RNG.
+* You should create one RNG at the beginning of your script (with a seed if you want reproducibility) and use this RNG in the rest of your script.
+
+To create a new RNG you can use the `default_rng` function as illustrated in the [introduction of the random module documentation](https://numpy.org/doc/1.18/reference/random/index.html#introduction):
+
+```python
+import numpy as np
+seed = 12345
```
tupui, line 37: Can you change that (and bellow) to a more random value? This is to prevent people that copy paste to 1. copy all the time the same value (hence arguably not have a RNG anymore and have the same number generator as everyone), 2. the doc suggest that larger numbers should be used to have a better entropy (though there are better safeguard now as opposed to the old infra as I understand it.)

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,110 @@
...
+
+### The main messages
+1. Avoid using the global numpy RNG. This means that you should avoid using [`np.random.seed`](https://numpy.org/doc/1.18/reference/random/generated/numpy.random.seed.html) and `np.random.*` functions, such as `np.random.random`, to generate random values.
+2. Create a new RNG and pass it around using the [`np.random.default_rng`](https://numpy.org/doc/1.18/reference/random/generator.html#numpy.random.default_rng) function.
+3. Be careful with parallel computations and rely on [numpy strategies for reproducible parallel number generation](https://numpy.org/doc/1.18/reference/random/parallel.html).
+
+Note that with numpy <1.17 the way to create a new RNG is to use [`np.random.RandomState`](https://numpy.org/doc/1.18/reference/random/legacy.html#numpy.random.RandomState) which is based on the popular Mersenne Twister 19937 algorithm. This is also how the global numpy RNG is created. It is still possible to use this function in versions higher than 1.17 but it is now recommended to use `default_rng` which returns an instance of the statistically better [PCG64](https://www.pcg-random.org) RNG.
```
tupui, line 20: It might be worth noting a slight nuance (either here or in some advance section or footnote). `np.random.RandomState` is going to be around and there is no plan to mark it at legacy either. This is pretty much the recommended RNG to use for testing as there are still the same guarantee as before in terms of byte stream stability. Which the new infra relaxes a bit.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,110 @@
...
+
+# rerun to check that we obtain the same outputs
+random_vector = Parallel(n_jobs=2)(delayed(
+    stochastic_function)(random_state) for random_state in child_states)
+print(random_vector)
+```
+
+By using a fixed seed you always get the same results and by using `SeedSequence.spawn` you have an independent RNG for each of the iterations. Note that I used the convenient `default_rng` function in `stochastic_function`. You can also see that the `SeedSequence` of the existing RNG is a private attribute. Accessing the `SeedSequence` might become easier in future versions of numpy (more information [here](https://github.com/numpy/numpy/issues/15322#issuecomment-626400433)).
+
+## Resources
```
tupui, line 100: Might be of interest, there are discussions still around RNG in the community to bring more uniformisation between packages. These are quite interesting discussions:

https://github.com/scientific-python/specs/pull/180

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,110 @@
...
+
+def stochastic_function(seed, high=10):
+    rng = np.random.default_rng(seed)
+    return rng.integers(high, size=5)
+
+seed = 98765
+# create the RNG that you want to pass around
+rng = np.random.default_rng(seed)
+# get the SeedSequence of the passed RNG
+ss = rng.bit_generator._seed_seq
```
tupui, line 83: I would add some very clear+visible warning that this is using a private API. In general we really try not to show usage of private methods as there is no guarantee that we won't change these (and we want to be free to do so).

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,110 @@
+---
+title: "Good practices with numpy random number generators"
```
tupui, line 2: When referring to the name of the project (so not code), we should use the proper capitalisation.
```suggestion
title: "Good practices with NumPy random number generators"
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,110 @@
...
+draft: false
+description: "
+Good practices with numpy random number generators
+"
+tags: ["tutorials", numpy, rng]
+displayInList: true
+author: ["Albert Thomas"]
+---
+
+Unless you are working on a problem where you can afford a true Random Number Generator (RNG), which is basically never for most of us, implementing something random means relying on a pseudo Random Number Generator. I want to share here what I have learned about good practices with pseudo RNGs and especially the ones available in [numpy](https://numpy.org/). I assume a certain knowledge of numpy and that numpy 1.17 or greater is used. The reason for this is that great new features were introduced in the [random](https://numpy.org/doc/1.18/reference/random/index.html) module of version 1.17. As `numpy` is usually imported as `np`, I will sometimes use `np` instead of `numpy`. Finally, as I will not talk about true RNGs, RNG will always mean pseudo RNG in the rest of this blog post.
```
tupui, line 13:
```suggestion
Unless you are working on a problem where you can afford a true Random Number Generator (RNG), which is basically never for most of us, implementing something random means relying on a pseudo Random Number Generator. I want to share here what I have learned about good practices with pseudo RNGs and especially the ones available in [NumPy](https://numpy.org/). I assume a certain knowledge of numpy and that NumPy 1.17 or greater is used. The reason for this is that great new features were introduced in the [random](https://numpy.org/doc/1.18/reference/random/index.html) module of version 1.17. As `numpy` is usually imported as `np`, I will sometimes use `np` instead of `numpy`. Finally, as I will not talk about true RNGs, RNG will always mean pseudo RNG in the rest of this blog post.
```

tupui (changes requested):
Thank you @albertcthomas for the contribution 🚀 This is really appreciated 😃 

I left a few suggestions, nothing major and I am positive to publish it.

stefanv:
@albertcthomas Thank you so much for this contribution! I have been on leave, but will review when I'm back.

content/posts/numpy/numpy-rng/index.md
albertcthomas: This is nice, I'll add this to the post.

content/posts/numpy/numpy-rng/index.md
albertcthomas: > though there are better safeguard now as opposed to the old infra as I understand it.

yes I thought that with the new seed sequence algorithm we were sure to get a good initial state for the RNG, even with a bad input seed from the user?

albertcthomas: > Can you change that (and bellow) to a more random value?

sorry can you elaborate a bit? you mean something that looks a bit more random like `74329087451`?

albertcthomas:
> @albertcthomas Thank you so much for this contribution! I have been on leave, but will review when I'm back.

No worries. I'm also a bit slow on this... Thanks @tupui for the great comments, it's very appreciated to get some feedbacks on this post, I'm learning additional stuffs while doing this. I'll integrate them ASAP.

content/posts/numpy/numpy-rng/index.md
albertcthomas: > This is pretty much the recommended RNG to use for testing as there are still the same guarantee as before in terms of byte stream stability

you mean for existing projects so that tests won't break?

content/posts/numpy/numpy-rng/index.md
albertcthomas: Ok I found more information about that at https://numpy.org/doc/stable/reference/random/index.html#random-sampling-numpy-random

content/posts/numpy/numpy-rng/index.md
tupui: Yes. But also if you write new tests, the official recommended way is to use `RandomState`. I agree that it's not clear from their doc, but maintainers of NumPy told me that a few times.

albertcthomas: is it that Mersenne Twister has better properties for testing?

albertcthomas:
> No worries. I'm also a bit slow on this...

I dealt with the comments made by @tupui. I will just re-read what I wrote after getting some sleep and push it :)

content/posts/numpy/numpy-rng/index.md
tupui: No the reason is that because they still provide strong guarantees of stability with `RandomState`, the tests are less likely to break. But because of known statistical issues, it's not recommended to use these for anything else than production. Bugs are known and cannot be fixed due to this same guarantee.

TBH, I personally only use `Generator` for all my code for simplicity. It's really not like this is going to break all the time.

albertcthomas:
By the way I would like to add one sentence to thank the reviewers that made this blog post better than it was (like @tupui and others if I got other reviews) and other people like @stefanv for suggesting putting my blog post here. Is that OK with the scientific python blog?

tupui:
Ok course, thank you 😃 This is really your post and we want to let people have their voice and write as their feel like. (And I hope we managed to do that here too 😅). We think it can be more interesting and engaging if we do as little editorial edits as possible to retain everyone's own intent, vision and ideas.

content/posts/numpy/numpy-rng/index.md
albertcthomas: Ok got it. I added a sentence, but do we want to discourage this kind of practice (even for tests) or at least not encourage it?

albertcthomas:
I think I addressed all your comments @tupui. Let me know if there is anything left. Thanks again!

albertcthomas:
> Ok course, thank you 😃 This is really your post and we want to let people have their voice and write as their feel like. (And I hope we managed to do that here too 😅). We think it can be more interesting and engaging if we do as little editorial edits as possible to retain everyone's own intent, vision and ideas.

I will add one then.

Also could you please let me know before merging this PR? I'd like to read it one last time before it goes live :)

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+2. Create a new RNG and pass it around using the [`np.random.default_rng`](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng) function.
+3. Be careful with parallel computations and rely on [NumPy strategies for reproducible parallel number generation](https://numpy.org/doc/stable/reference/random/parallel.html).
+
+Note that with NumPy <1.17 the way to create a new RNG is to use [`np.random.RandomState`](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) which is based on the popular Mersenne Twister 19937 algorithm. This is also how the global NumPy RNG is created. It is still possible to use this function in versions higher than 1.17 but it is now recommended to use `default_rng` which returns an instance of the statistically better [PCG64](https://www.pcg-random.org) RNG. You might see `np.random.RandomState` being used in tests as it has strong stability guarantees between different NumPy versions.
```
stefanv, line 20:
```suggestion
Note that, with the older NumPy <1.17, the way to create a new RNG is to use [`np.random.RandomState`](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) which is based on the popular Mersenne Twister 19937 algorithm. This is also how the global NumPy RNG is created. This function is still available in newer versions of NumPy, but it is now recommended to use `default_rng` instead, which returns an instance of the statistically better [PCG64](https://www.pcg-random.org) RNG. You might still see `np.random.RandomState` being used in tests as it has strong stability guarantees between different NumPy versions.
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Random number generation with NumPy
+When you import `numpy` in your python script a RNG is created behind the scenes. This RNG is the one used when you generate a new random value using a function such as `np.random.random`. I will here refer to this RNG as the global NumPy RNG.
```
stefanv, line 23:
```suggestion
When you import `numpy` in your python script, a RNG is created behind the scenes. This RNG is the one used when you generate a new random value using a function such as `np.random.random`. I will here refer to this RNG as the global NumPy RNG.
```

I would probably use "an RNG", but "a RNG" is OK if you imagine people reading "a random number generator".

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+Although not recommended, it is a common practice to reset the seed of this global RNG at the beginning of a script using the `np.random.seed` function. Fixing the seed at the beginning ensures that the script is reproducible: the same values and results will be produced each time you run it. However, although sometimes convenient, using the global NumPy RNG is considered a bad practice. A simple reason is that using global variables can lead to undesired side effects. For instance one might use `np.random.random` without knowing that the seed of the global RNG was set somewhere else in the codebase. Quoting the [Numpy Enhancement Proposal (NEP) 19](https://numpy.org/neps/nep-0019-rng-policy.html) by Robert Kern about the NumPy RNG policy:
```
stefanv, line 25:
```suggestion
Although not recommended, it is a common practice to reset the seed of this global RNG at the beginning of a script using the `np.random.seed` function. Fixing the seed at the beginning ensures that the script is reproducible: the same values and results will be produced each time you run it. However, although sometimes convenient, using the global NumPy RNG is considered a bad practice. A simple reason is that using global variables can lead to undesired side effects. For instance one might use `np.random.random` without knowing that the seed of the global RNG was set somewhere else in the codebase. Quoting [Numpy Enhancement Proposal (NEP) 19](https://numpy.org/neps/nep-0019-rng-policy.html) by Robert Kern:
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+> The implicit global RandomState behind the `np.random.*` convenience functions can cause problems, especially when threads or other forms of concurrency are involved. Global state is always problematic. We categorically recommend avoiding using the convenience functions when reproducibility is involved. [...] The preferred best practice for getting reproducible pseudorandom numbers is to instantiate a generator object with a seed and pass it around.
+
+In short:
+* Instead of using `np.random.seed`, which reseeds the already created global NumPy RNG and then using `np.random.*` functions you should create a new RNG.
```
stefanv, line 30:
```suggestion
* Instead of using `np.random.seed`, which reseeds the already created global NumPy RNG, and then using `np.random.*` functions, you should create a new RNG.
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+* You should create one RNG at the beginning of your script (with a seed if you want reproducibility) and use this RNG in the rest of your script.
```
stefanv, line 31:
```suggestion
* You should create an RNG at the beginning of your script (with a seed, if you want reproducibility) and use this RNG in the rest of the script.
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+To create a new RNG you can use the `default_rng` function as illustrated in the [introduction of the random module documentation](https://numpy.org/doc/stable/reference/random/index.html):
+
+```python
+import numpy as np
+
+rng = np.random.default_rng()
+rng.random()
```
stefanv, line 39:
```suggestion
rng.random()  # generate a floating point number between 0 and 1
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+If you want to use a seed for reproducibility, [Numpy documentation](https://numpy.org/doc/stable/reference/random/index.html#quick-start) recommends using a very large and unique number to have a different seed than anyone else. Another reason is that using a small number for the seed could lead to a bad initialization of the RNG, although `default_rng` now takes care of creating a good initialization from a small number. To obtain a good unique seed you can rely on the [secrets module](https://docs.python.org/3/library/secrets.html). The following code will return a 128 bit number that you can then use as a seed:
```
stefanv, line 42:
```suggestion
If you want to use a seed for reproducibility, [the NumPy documentation recommends](https://numpy.org/doc/stable/reference/random/index.html#quick-start) using a very large and unique number to have a different seed than anyone else. Another reason is that using a small number for the seed could lead to a bad initialization of the RNG, although `default_rng` now takes care of creating a good initialization from a small number. To obtain a good, unique seed you can rely on the [secrets module](https://docs.python.org/3/library/secrets.html). The following code will return a 128 bit number that you can then use as a seed:
```

stefanv, line 42: Maybe worth explaining what a seed is somewhere higher up?

stefanv: Most people may not see why it's necessary to have different random numbers than other people.

stefanv: What does "bad initialization" mean to the reader?

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+rng = np.random.default_rng(seed)
+rng.random()
+```
+
+The reason for seeding your RNG only once (and passing that RNG around) is that you can lose on the randomness and the independence of the generated random numbers by reseeding the RNG multiple times. Furthermore obtaining a good seed can be time consuming. Once you have a good seed to instantiate your generator you might as well use it. With a good RNG such as the one returned by `default_rng` you will be ensured good randomness (and independence) of the generated numbers. It might be more dangerous to use different seeds: how do you know that the streams of random numbers obtained with two different seeds are not correlated, or I should say less independent than the ones created from the same seed? That being said, [as explained by Robert Kern](https://github.com/numpy/numpy/issues/15322#issuecomment-573890207), with the RNGs and seeding strategies introduced in NumPy 1.17, it could be considered safe enough to recreate new RNGs from the system entropy, e.g. using `default_rng(None)` multiple times. However as explained later be careful when running jobs in parallel and relying on `default_rng(None)`.
```
stefanv, line 60: Can you clarify "lose on the randomness and the independence"?

stefanv, line 60:
```suggestion
The reason for seeding your RNG only once (and passing that RNG around) is that you can lose on the randomness and the independence of the generated random numbers by reseeding the RNG multiple times. Furthermore obtaining a good seed can be time consuming. Once you have a good seed to instantiate your generator, you might as well use it. With a good RNG such as the one returned by `default_rng` you will be ensured good randomness (and independence) of the generated numbers. It might be more dangerous to use generators derived from different seeds: how do you know that the streams of random numbers obtained from those are not correlated or, I should say, less independent than the ones created from the same seed? [As explained by Robert Kern](https://github.com/numpy/numpy/issues/15322#issuecomment-573890207), with the RNGs and seeding strategies introduced in NumPy 1.17, it is considered fairly safe to create RNGs using system entropy, i.e. using `default_rng(None)` multiple times. However as explained later be careful when running jobs in parallel and relying on `default_rng(None)`.
```

stefanv: It's a bit unclear how the sentences on `default_rng(None)` relate to seeding. Above, you say not to use `None` as the input to `default_rng`, and now that is being suggested, so would be worth clarifying when that's a good idea.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+As you write functions that you will use on their own as well as in a more complex script it is convenient to be able to pass a seed or your already created RNG. The function `default_rng` allows you to do this very easily. As written above, this function can be used to create a new RNG from your chosen seed, if you pass a seed to it, or from system entropy when passing `None` but you can also pass an already created RNG. In this case the returned RNG is the one that you passed.
+
+```python
+import numpy as np
+
+
+def stochastic_function(seed, high=10):
+    rng = np.random.default_rng(seed)
+    return rng.integers(high, size=5)
```
stefanv, line 73: The API we've been converging on in SPEC 7 is to call the argument `rng` instead of `seed`, but it's fine this way too.

See https://github.com/scientific-python/specs/pull/180

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+Before knowing about `default_rng`, and before NumPy 1.17, I was using the scikit-learn function [`check_random_state`](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html) which is of course heavily used in the scikit-learn codebase. While writing this post I discovered that this function is now available in [scipy](https://github.com/scipy/scipy/blob/master/scipy/_lib/_util.py#L171). A look at the docstring and/or the source code of this function will give you a good idea about what it does. The differences with `default_rng` are that `check_random_state` currently relies on `np.random.RandomState` and that when `None` is passed to `check_random_state` then the function returns the already existing global NumPy RNG. The latter can be convenient because if you fix the seed of the global RNG before in your script using `np.random.seed`, `check_random_state` returns the generator that you seeded. However, as explained above, this is not the recommended practice and you should be aware of the risks and the side effects.
```
stefanv, line 77: Perhaps better as a footnote, so as not to distract the reader.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Parallel processing
+
+You must be careful when using RNGs in conjunction with parallel processing. I usually use the [joblib](https://joblib.readthedocs.io/en/latest/) library to parallelize my code and I will therefore mainly talk about it. However most of the discussion is not specific to joblib.
```
stefanv, line 81: The "I usually use", "However..." sentences distract a bit from the gist of what you want to convey here. Perhaps simply start with: "You must be careful when using RNGs in parallel processing"... <and here's why> ... and then add any technical details about the libraries you will use/discuss.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+If you fix the seed at the beginning of your main script for reproducibility and then pass the same RNG to each process to be run in parallel, most of the time this will not give you what you want as this RNG will be deep copied. The same results will thus be produced by each process. One of the solutions is to create as many RNGs as parallel processes with a different seed for each of these RNGs. The issue now is that you cannot choose the seeds as easily as you would think. When you choose two different seeds to instantiate two different RNGs how do you know that the numbers produced by these RNGs will appear as statistically independent? The design of independent RNGs for parallel processes has been an important research question. You can for instance refer to the paper [Random numbers for parallel computers: Requirements and methods, with emphasis on GPUs](https://www.sciencedirect.com/science/article/pii/S0378475416300829) by L'Ecuyer et al. (2017) for a good summary of the different methods on this topic.
```
stefanv, line 85: "reused" instead of "deepcopied"? Each time we use jargon, it risks confusing those not familiar with it, and I don't think the exact mechanism is too relevant here.

stefanv, line 85:
```suggestion
If you fix the seed at the beginning of your main script for reproducibility and then pass the same RNG to each process to be run in parallel, most of the time this will not give you what you want as this RNG will be deep copied. The same results will thus be produced by each process. One of the solutions is to create as many RNGs as parallel processes with a different seed for each of these RNGs. The issue now is that you cannot choose the seeds as easily as you would think. When you choose two different seeds to instantiate two different RNGs how do you know that the numbers produced by these RNGs will appear as statistically independent? The design of independent RNGs for parallel processes has been an important research question. See, for example, [Random numbers for parallel computers: Requirements and methods, with emphasis on GPUs](https://www.sciencedirect.com/science/article/pii/S0378475416300829) by L'Ecuyer et al. (2017) for a good summary of different methods.
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+From NumPy 1.17, it is now very easy to instantiate independent RNGs. Depending on the RNG that you use, different strategies are easily available as documented in the [Parallel random number generation section](https://numpy.org/doc/stable/reference/random/parallel.html) of the NumPy documentation. One of the strategies is to use `SeedSequence` which is an algorithm that makes sure that a not so good user-provided seed results in a good initial state for the RNG. Additionally, it ensures that two close seeds will result in two very different initial states for the RNG that are, with very high probability, independent of each other. You can refer to the documentation of [SeedSequence Spawning](https://numpy.org/doc/stable/reference/random/parallel.html#seedsequence-spawning) for an example on how to generate independent RNGs from a user-provided seed. I here show how you can do this from an already existing RNG and apply it to the [joblib example](https://joblib.readthedocs.io/en/latest/auto_examples/parallel_random_state.html#fixing-the-random-state-to-obtain-deterministic-results) mentioned above.
```
stefanv, line 87:
```suggestion
Starting with NumPy 1.17, it is now very easy to instantiate independent RNGs. Depending on the type of RNG you use, different strategies are available as documented in the [Parallel random number generation section](https://numpy.org/doc/stable/reference/random/parallel.html) of the NumPy documentation. One of the strategies is to use `SeedSequence` which is an algorithm that makes sure that poor input seeds are transformed into good initial RNG states. Additionally, it ensures that close seeds are mapped to very different initial states, resulting in RNGs that are, with very high probability, independent of each other. You can refer to the documentation of [SeedSequence Spawning](https://numpy.org/doc/stable/reference/random/parallel.html#seedsequence-spawning) for an example on how to generate independent RNGs from a user-provided seed. I here show how you can do this from an already existing RNG and apply it to the [joblib example](https://joblib.readthedocs.io/en/latest/auto_examples/parallel_random_state.html#fixing-the-random-state-to-obtain-deterministic-results) mentioned above.
```

content/posts/numpy/numpy-rng/index.md
stefanv: Why not start from a `SeedSequence`?

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+# rerun to check that we obtain the same outputs
+random_vector = Parallel(n_jobs=2)(
+    delayed(stochastic_function)(random_state) for random_state in child_states
+)
+print(random_vector)
+```
+
+By using a fixed seed you always get the same results and by using `SeedSequence.spawn` you have an independent RNG for each of the iterations. Note that I used the convenient `default_rng` function in `stochastic_function`. You can also see that the `SeedSequence` of the existing RNG is a private attribute. Accessing the `SeedSequence` might become easier in future versions of NumPy (more information [here](https://github.com/numpy/numpy/issues/15322#issuecomment-626400433)).
+
```
stefanv, line 124: The end feels a bit abrupt; it may be helpful to have a short conclusion here. It could simply say that the new random API provides safer ways to generate near-independent random streams of numbers, and that the resources below are available for further reading.

stefanv (commented):
@albertcthomas Thanks again for the blog post! I've done a quick review; feel free to use what you want from the comments and disregard the rest.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+draft: false
+description: "
+Good practices with NumPy random number generators
+"
+tags: ["tutorials", numpy, rng]
+displayInList: true
+author: ["Albert Thomas"]
+---
+
+Unless you are working on a problem where you can afford a true Random Number Generator (RNG), which is basically never for most of us, implementing something random means relying on a pseudo Random Number Generator. I want to share here what I have learned about good practices with pseudo RNGs and especially the ones available in [NumPy](https://numpy.org/). I assume a certain knowledge of NumPy and that NumPy 1.17 or greater is used. The reason for this is that great new features were introduced in the [random](https://numpy.org/doc/stable/reference/random/index.html) module of version 1.17. As `numpy` is usually imported as `np`, I will sometimes use `np` instead of `numpy`. Finally, as I will not talk about true RNGs, RNG will always mean pseudo RNG in the rest of this blog post.
```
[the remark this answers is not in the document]
jarrodmillman: I suspect many theoretical computer scientists would take this view:
```
Using Physical Random Sources: It is unclear whether the real world has physical sources
of perfect randomness. We may use sources that seem to have some unpredictability, like
the low order bits of a system clock or thermal noise, but these sources will generally
have biases and, more problematically, correlations. Thus we ask: What can we do with
a source of biased and correlated bits?
```
https://people.seas.harvard.edu/~salil/pseudorandomness/pseudorandomness-Aug10.pdf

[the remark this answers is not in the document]
albertcthomas: Thanks for the ref @jarrodmillman I was not really aware of these issues about True RNGs

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+### The main messages
+1. Avoid using the global NumPy RNG. This means that you should avoid using [`np.random.seed`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) and `np.random.*` functions, such as `np.random.random`, to generate random values.
+2. Create a new RNG and pass it around using the [`np.random.default_rng`](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng) function.
```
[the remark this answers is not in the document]
albertcthomas: Sorry I will revert this suggested commit, the `np.random.default_rng` is used to create the RNG as well as passing it around so I'd prefer keeping the original formulation.

content/posts/numpy/numpy-rng/index.md
albertcthomas: > What does "bad initialization" mean to the reader?

good point, the issue here is that I do not want to enter into the details of how an RNG works and talk about the RNG states etc... If people want to know more they can check the references given at the end, but I agree that bad initialization is not very clear. I could probably remove this, given that the new SeedSequence algorithm is supposed to build a good initial state from any seed.

albertcthomas: not sure I want to enter into this, I suspect the reader to be familiar with seeding and this is a guide on how to do this properly. Or I would need to write a small paragraph in the intro explaining a bit how an RNG works.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Random number generation with NumPy
+When you import `numpy` in your python script, an RNG is created behind the scenes. This RNG is the one used when you generate a new random value using a function such as `np.random.random`. I will here refer to this RNG as the global NumPy RNG.
+
+Although not recommended, it is a common practice to reset the seed of this global RNG at the beginning of a script using the `np.random.seed` function. Fixing the seed at the beginning ensures that the script is reproducible: the same values and results will be produced each time you run it. However, although sometimes convenient, using the global NumPy RNG is considered a bad practice. A simple reason is that using global variables can lead to undesired side effects. For instance one might use `np.random.random` without knowing that the seed of the global RNG was set somewhere else in the codebase. Quoting [Numpy Enhancement Proposal (NEP) 19](https://numpy.org/neps/nep-0019-rng-policy.html) by Robert Kern:
```
jarrodmillman, line 25:
```suggestion
Although not recommended, it is a common practice to reset the seed of this global RNG at the beginning of a script using the `np.random.seed` function. Fixing the seed at the beginning ensures that the script is reproducible: the same values and results will be produced each time you run it. However, although sometimes convenient, using the global NumPy RNG is a bad practice. A simple reason is that using global variables can lead to undesired side effects. For instance one might use `np.random.random` without knowing that the seed of the global RNG was set somewhere else in the codebase. Quoting [Numpy Enhancement Proposal (NEP) 19](https://numpy.org/neps/nep-0019-rng-policy.html) by Robert Kern:
```

content/posts/numpy/numpy-rng/index.md
albertcthomas: for this example I agree that we could start from a `SeedSequence` but when this occurs at any place in a larger python code, and you are only given the passed RNG you need to get the seed sequence of that RNG.

albertcthomas:
Thanks a lot for the new reviews, I committed the easy suggestions, I will incorporate more of them later.

content/posts/numpy/numpy-rng/index.md
albertcthomas: yes good point again, what's not clear in my text is that `default_rng(None)` will still need to obtain a seed from somewhere (e.g. system entropy)

content/posts/numpy/numpy-rng/index.md
albertcthomas: On second thoughts, I feel deep copied is more clear as this means that the RNGs will have the same state in each process and thus produce the same numbers. I think reused is a bit less explicit here.

content/posts/numpy/numpy-rng/index.md
albertcthomas: I think I will add a small paragraph in the intro to clarify all this.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Passing a NumPy RNG around
+
+As you write functions that you will use on their own as well as in a more complex script it is convenient to be able to pass a seed or your already created RNG. The function `default_rng` allows you to do this very easily. As written above, this function can be used to create a new RNG from your chosen seed, if you pass a seed to it, or from system entropy when passing `None` but you can also pass an already created RNG. In this case the returned RNG is the one that you passed.
+
+```python
+import numpy as np
+
+
+def stochastic_function(seed, high=10):
```
albertcthomas, line 71: note to myself: seed -> rng

albertcthomas:
Fixed the pre-commit!

tupui (approved):
This is great! Thank you for the update @albertcthomas. For the date yes I think it's better to update 😅 

I read again the article and I don't have any more suggestions 😃 

@stefanv @jarrodmillman does that look good for you as well?

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,144 @@
...
+3. Be careful with parallel random number generation and use the [strategies provided by NumPy](https://numpy.org/doc/stable/reference/random/parallel.html).
+
+Note that, with the older NumPy <1.17, the way to create a new RNG is to use [`np.random.RandomState`](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) which is based on the popular Mersenne Twister 19937 algorithm. This is also how the global NumPy RNG is created. This function is still available in newer versions of NumPy, but it is now recommended to use `default_rng` instead, which returns an instance of the statistically better [PCG64](https://www.pcg-random.org) RNG. You might still see `np.random.RandomState` being used in tests as it has strong stability guarantees between different NumPy versions.
+
+## Random number generation with NumPy
+
+When you import `numpy` in your python script, an RNG is created behind the scenes. This RNG is the one used when you generate a new random value using a function such as `np.random.random`. I will here refer to this RNG as the global NumPy RNG.
```
stefanv, line 29:
```suggestion
When you import `numpy` in your Python script, an RNG is created behind the scenes. This RNG is the one used when you generate a new random value using a function such as `np.random.random`. I will here refer to this RNG as the global NumPy RNG.
```

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,144 @@
...
+rng = np.random.default_rng()
+rng.random()  # generate a floating point number between 0 and 1
+```
+
+If you want to use a seed for reproducibility, [the NumPy documentation recommends](https://numpy.org/doc/stable/reference/random/index.html#quick-start) using a very large and unique number to have a different seed than anyone else. Another reason is that using a small number for the seed could lead to a bad first internal state of the RNG leading to a sequence of numbers that would not look like a random sequence, although `default_rng` now takes care of creating a good internal state from a small number. To obtain a good and unique seed you can rely on the [secrets module](https://docs.python.org/3/library/secrets.html). The following code will return a 128 bit number that you can then use as a seed:
+
+```python
+import secrets
+
+secrets.randbits(128)
```
stefanv, line 54: Given that `default_rng` does a good job of seeding from a small number, will 64 suffice? 32?

I suppose it does not matter whether you start with a small or large number, if all you want is reproducibility. Is the concern with small seeds not what happens when you instantiate multiple PRNGs, using x, x+1, x+2, etc?

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+```python
+import numpy as np
+
+seed = 65647437836358831880808032086803839626
+rng = np.random.default_rng(seed)
+rng.random()
+```
+
+The reason for seeding your RNG only once (and passing that RNG around) is that with a good RNG such as the one returned by `default_rng` you will be ensured good randomness and independence of the generated numbers. However, if not done properly, using several RNGs (each one created with its own seed) might lead to streams of random numbers that are less independent than the ones created from the same seed. That being said, [as explained by Robert Kern](https://github.com/numpy/numpy/issues/15322#issuecomment-573890207), with the RNGs and seeding strategies introduced in NumPy 1.17, it is considered fairly safe to create RNGs using system entropy, i.e. using `default_rng(None)` multiple times. However as explained later be careful when running jobs in parallel and relying on `default_rng(None)`. Another reason for seeding your RNG only once is that obtaining a good seed can be time consuming. Once you have a good seed to instantiate your generator, you might as well use it.
```
stefanv, line 67: Independence of generated numbers is a property of the algorithm, for any given seed?

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Passing a NumPy RNG around
+
+As you write functions that you will use on their own as well as in a more complex script it is convenient to be able to pass a seed or your already created RNG. The function `default_rng` allows you to do this very easily. As written above, this function can be used to create a new RNG from your chosen seed, if you pass a seed to it, or from system entropy when passing `None` but you can also pass an already created RNG. In this case the returned RNG is the one that you passed.
+
+```python
+import numpy as np
+
+
+def stochastic_function(rng, high=10):
```
stefanv, line 77: Typical signature would be `stochastic_function(high=10, rng=None)` or similar?

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,144 @@
...
+# rerun to check that we obtain the same outputs
+random_vector = Parallel(n_jobs=2)(
+    delayed(stochastic_function)(random_state) for random_state in child_states
+)
+print(random_vector)
+```
+
+By using a fixed seed you always get the same results and by using `SeedSequence.spawn` you have an independent RNG for each of the iterations. Note that I used the convenient `default_rng` function in `stochastic_function`. You can also see that the `SeedSequence` of the existing RNG is a private attribute. Accessing the `SeedSequence` might become easier in future versions of NumPy (more information [here](https://github.com/numpy/numpy/issues/15322#issuecomment-626400433)).
+
+I hope this blog post helped you understand the best ways to use Numpy RNGs. The new Numpy API gives you all the tools you need for that. The resources below are available for further reading. Finally, I would like to thank Pamphile Roy, Stefan van der Walt and Jarrod Millman for their great feedbacks and comments which contributed to greatly improve the original version of this blog post.
```
stefanv, line 129:
```suggestion
I hope this blog post helped you understand the best ways to use NumPy RNGs. The new Numpy API gives you all the tools you need for that. The resources below are available for further reading. Finally, I would like to thank Pamphile Roy, Stefan van der Walt and Jarrod Millman for their great feedbacks and comments which contributed to greatly improve the original version of this blog post.
```

stefanv (commented):
Thanks @albertcthomas! This looks great. I left another round of comments, but at any point I am happy for you to suggest that we merge (and when you do so, I will!).

content/posts/numpy/numpy-rng/index.md
albertcthomas: Yes true I'll fix this

content/posts/numpy/numpy-rng/index.md
```diff
@@ -0,0 +1,144 @@
...
+draft: false
+description: "
+Good practices with NumPy random number generators
+"
+tags: ["tutorials", numpy, rng]
+displayInList: true
+author: ["Albert Thomas"]
+---
+
+Unless you are working on a problem where you can afford a true Random Number Generator (RNG), which is basically never for most of us, implementing something random means relying on a pseudo Random Number Generator. I want to share here what I have learned about good practices with pseudo RNGs and especially the ones available in [NumPy](https://numpy.org/).
```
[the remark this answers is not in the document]
albertcthomas: Well it seems that they exist according to a google search (e.g. https://en.wikipedia.org/wiki/Hardware_random_number_generator) but I agree that the term can be misleading and that they have their own issues. I just wanted to say here that PRNGs are not the only way to get random numbers but I will remove this.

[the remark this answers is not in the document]
stefanv: Yeah, it's a bit weird:

> While "classical" (non-quantum) phenomena are not truly random, an unpredictable physical system is usually acceptable as a source of randomness, so the qualifiers "true" and "physical" are used interchangeably.

In that case, one could consider /dev/urandom truly random as well, since it's connected to a piece of hardware.

content/posts/numpy/numpy-rng/index.md
albertcthomas: From my understanding using a large number prevents you (with higher probability) from using the same as someone else. So someone would be able to check if they can confirm the results of a study I did without using the same seed by chance. But please correct me if this not why we would want a different seed than anyone else?

The concern is the same for small and large seeds when you instantiate multiple RNGs using x, x+1 etc... you do not know if your seeds will lead to less independent numbers although again default_rng is supposed to do a good job at creating good initial states from close seeds.

content/posts/numpy/numpy-rng/index.md
albertcthomas: Yes for a given seed, the algorithm produces independent random numbers. For two different seeds you do not know whether the number generated by the different seeds conserve this property. For instance your second seed could lead to the internal state that you obtain after 3 calls to the RNG instantiated with the first seeds. Then you will have an overlap of the values generated by the two RNGs. Do you want me to clarify this?

content/posts/numpy/numpy-rng/index.md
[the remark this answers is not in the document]
albertcthomas: yes I learned something :)

content/posts/numpy/numpy-rng/index.md
stefanv: You are quite right. So it depends on whether you want independence from others (or subsequent runs of your own), or independence between streams. Often, we only want reproducibility.

content/posts/numpy/numpy-rng/index.md
stefanv: Personally, I find these subtleties interesting. Perhaps footnotes? But I leave it up to you.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -1,18 +1,18 @@
...
+A pseudo RNG works by updating an internal state through a deterministic algorithm. The internal state is initialized with a value known as a seed and each update of this internal state produces a number that appears randomly generated. The key here is that the process is deterministic, meaning that if you start with the same seed and apply the same algorithm, you will get the same sequence of internal states (and numbers). Despite this determinism, the resulting numbers exhibit properties of randomness, appearing unpredictable and evenly distributed. Users can either specify the seed manually, providing a degree of control over the generated sequence, or they can opt to let the RNG object automatically derive the seed from system entropy. The latter approach enhances unpredictability by incorporating external factors into the seed.
```
stefanv, line 15: Remember to say what RNG stands for and what it is, before diving in

albertcthomas: indeed, done :)

albertcthomas:
OK to merge for me if you have no further comments :). Thanks again for the constructive comments

stefanv:
Excellent, I think we're good to go! Thanks so much @albertcthomas.

stefanv:
In it goes! Thanks so much, Albert! 🎉

albertcthomas:
Great! thanks a lot for reaching out and the reviews :)

content/posts/numpy/numpy-rng/index.md
seberg: I am not sure why the use of private API here.  I think this was public for a long time (although maybe necessary for the legacy mtrand?)

We also implemnted `rng.spawn(1)` and `bitgen.spawn(1)` more recently, although I think allowing `child = rng.spawn()` (no number to get a single one) might be a nice extension.

albertcthomas: thanks @seberg for the suggestion! I was not aware of the spawn method, just saw it's in numpy >=1.25, good to know!

I checked for the _seed_seq attribute quickly in the doc but could not find anything (https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.bit_generator.html#numpy.random.Generator.bit_generator), am I missing something?

While you are I actually have another question regarding seeding: it's written in the doc of the random module that it is good to have a large number as a seed to have a unique number different than anyone else. But it is also said in the bit generator doc that a seed lower than 32 bits means that there are some initial states that are impossible to reach. Is that the case even with seed sequence? I thought seed sequence was supposed to create good initial states from bad seed but as I saw on the PCG website I guess seed sequence cannot do everything :).

@stefanv I can open another PR to improve this if that's possible?

stefanv:
@albertcthomas You are welcome to modify your blog post!

content/posts/numpy/numpy-rng/index.md
stefanv: @albertcthomas You are welcome to modify your blog post!

seberg: > ed lower than 32 bits means that there are some initial states that are impossible to reach. Is that the case even with seed sequence? I thought seed sequence was supposed to create good initial states from bad seed but as I saw on the PCG website I guess seed sequence cannot do everything :).

I am not the person who knows such details without research, Robert Kern would be.  Seedsequence mixes well, maybe.  But if the initial states has more than 32bit of space, then you need more than 32bits of state to reach it all, maybe?

FWIW, I sometimes use:
```
secrets.randbits(128)
```
to just find a perfect seed in the docs (can throw in a `hex` if you like).

PR #143, opened 2023-10-11
Update hugo (dark mode)

jarrodmillman:
There are several improvements/changes, but the main one is support for dark mode. There is no selector currently. It uses whatever your browser settings have selected. It also got rid of some empty white space, so I increased the number of posts displayed on the front page from 5 to 7.

### DIFF
--- a/layouts/partials/section.html
+++ b/layouts/partials/section.html
@@ -3,7 +3,7 @@
 Recent Posts
 <div class="content-container">
   <!-- this section is populated by pulling posts from the site -->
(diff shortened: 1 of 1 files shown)


jarrodmillman:
## Before

2023-10-11T15:19:28,095419027-07:00

## After

2023-10-11T15:19:42,875712644-07:00

2023-10-11T15:22:06,013753498-07:00

tupui (approved):
Awesome 🎉

PR #148, opened 2023-10-14
Simplify two-column layout

stefanv:
Instead of grouping posts in groups of HTML elements, use CSS: a wrapping flex container.

Also change the "All posts" button text to "Browse all posts".


jarrodmillman (approved):

alphapapa:
Great!

PR #150, opened 2023-10-24
Suppress summaries on all posts

stefanv:
Suppress summaries on All Posts page

Previously, the theme was modified to support the noSummary parameter.
However, that change modifed the context in which the post_meta.html was
rendered, breaking other sites.

Here, we implement that functionality using ".Scratch.Set" to
set a temporary variable on the Page context instead.


jarrodmillman:
What should we do with the summaries now? They often have a lot of overlap with the first paragraph of the blog post. For example,

2023-10-24T13:21:37,463391678-07:00

But most of the posts that I checked have fairly redundant text now. Should we delete all the descriptions, just not show them anywhere, or ask all the original authors to rewrite either their summary or their first paragraph?

jarrodmillman:
We may also need to change `archetypes/default.md` depending on what you decide. We should also scan the instructions to be sure they are up to date with these changes.

stefanv:
Hrm, summary should definitely not display on post page! I'll take a look.

jarrodmillman:
If you don't want to display the summaries on the "all posts page", I wonder if you could have the summary show when you hover of something? We could also just not display them for now, then open an issue to revisit this, and merge this PR so the theme CI can pass.

jarrodmillman (approved):

PR #154, opened 2023-10-26
Add comments via giscus

### LINKED ISSUE #113: Add comments system
Adding comments will allow us to publish slightly more controversial opinion pieces, since the community will then be able to add different perspectives.

Consider public hosted tools such as (update list as you find options):

- https://github.com/giscus/giscus
- https://github.com/utterance/utterances (not sure it's being maintained?

See also https://gohugo.io/content-management/comments/

stefanv:
Closes #113

The one change we may want to consider, from a security viewpoint, is vendoring the Javascript.

### DIFF
--- a/layouts/partials/single/content.html
+++ b/layouts/partials/single/content.html
@@ -1,0 +1,43 @@
+<section class="content-padding flex-row">
+    <div class="shortcuts-container">
+        <div><i class="fa-solid fa-list"></i> On this page</div>
+        <div id="shortcuts"></div>
+    </div>
+    <div class="content-container">
+        <div class="columns">
+            <div class="column is-centered-tablet-portrait">
+                <h1 class="title section-title">{{ .Title }}</h1>
+                <h5 class="subtitle is-5 is-muted">{{ .Params.Subtitle }}</h5>
+                <div class="divider"></div>
+            </div>
+        </div>
+        {{ with partial "post_meta.html" . -}}
+        <div class="post-meta">{{ . }}</div>
+        {{- end }}
+
+        <div class="post-content">
+          {{ .Content }}
+        </div>
+
+        <div class="giscus">
+        </div>
+
+        <script src="https://giscus.app/client.js"
+                data-repo="scientific-python/blog.scientific-python.org-comments"
+                data-repo-id="R_kgDOKlLVsA"
+                data-category="Comments"
+                data-category-id="DIC_kwDOKlLVsM4CacnO"
+                data-mapping="pathname"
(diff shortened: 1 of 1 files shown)


stefanv:
See https://deploy-preview-154--blog-scientific-python-org.netlify.app/scientific-python/dev-summit-1-development-guide/ for an example comment.

tupui (approved):
This looks great 👍 

If we vendor we need to also add that to dependabot somehow. Not sure it's worth it.

Now the major change this introduces is that we take user inputs. Do we fully trust this mechanism? Is it hardened enough?

stefanv:
It has 6k stars on GH. Maybe we try it and take it out if it's a mess?

tupui:
All good with me 😃 I was just curious if you had more insights.

stefanv:
No, sorry :/ I was going to use utterance, but then noticed that it hadn't been worked on in years, so this seemed like a better alternative. It is also listed as one of the top options in Hugo's docs.

tupui:
If Hugo recommends it then let's go 👍

stefanv:
Moving this topic to the theme.

stefanv:
Maybe better to get something in, rather than nothing, for now?

jarrodmillman:
How hard would it be to use discourse? It would be good to have all the discussions in one place and it could increase discussions on discourse.

stefanv:
It's not that hard. We just have to embed an iframe:

https://blog.codinghorror.com/welcome-to-the-internet-of-compromised-things/#discourse-comments

However, only people with Discourse accounts will be able to comment. If that's okay, I can get this implemented right away.

stefanv:
After an unreasonable amount of hacking and Javascript, I have something working.

This PR implements giscus, which looks like this:



The Discus embed renders like this in light mode:



And (the part that was fairly tricky to make work, and for which I haven't selected all the colors yet) in dark mode:



As you can see, no code highlight in Discourse, and no comment entry box (for that you have to navigate to https://discuss.scientific-python.org).

Comments are [here](https://github.com/scientific-python/blog.scientific-python.org-comments/discussions/1#discussioncomment-7386978) and [here](https://discuss.scientific-python.org/t/scientific-python-blog-the-scientific-python-development-guide/901/2?u=stefanv), respectively.

Let me know how I should proceed?

stefanv:
Page without comments: https://deploy-preview-177--blog-scientific-python-org.netlify.app/numpy/numpy-rng/

stefanv:
We merged #177 instead.

PR #166, opened 2023-11-07
Update theme

stefanv:


jarrodmillman:
See #164. There is more cleanup that needs to go into this and I was planning to do it with a designated release, which I will do tonight.

PR #179, opened 2024-02-12
Update numpy RNG post

### LINKED ISSUE #176: Question about Numpy random numbers blog post
When using the random state for parallel processing, do we need the private API?

https://github.com/scientific-python/blog.scientific-python.org/blob/b51e4b9df0ea5fd740b0f04c43ab432388deed21/content/posts/numpy/numpy-rng/index.md?plain=1#L108

From my understanding, we can use the public API:

```python
import numpy as np
from joblib import Parallel, delayed


def stochastic_function(high=10, rng=None):
    if not isinstance(rng, np.random.Generator):
        rng = np.random.default_rng(rng)
    return rng.integers(high, size=5)


seed = 319929794527176038403653493598663843656
rng = np.random.default_rng(seed)
child_states = rng.spawn(5)

random_vector = Parallel(n_jobs=2)(
    delayed(stochastic_function)(rng=random_state) for random_state in child_states
)
print(random_vector)

random_vector = Parallel(n_jobs=2)(
    delayed(stochastic_function)(rng=random_state) for random_state in child_states
)
print(random_vector)
```

CC @albertcthomas

albertcthomas:
Fixes #176 . See also https://github.com/scientific-python/blog.scientific-python.org/pull/135/files#r1470002426

This updates the way to spawn from the RNG using numpy 1.25.

I also added some information on seeding.

I have a weird formatting fix from prettier for the footnotes. I'll check the rendering.


albertcthomas:
Hm not sure why pre-commit complains about an unrelated file.

stefanv:
pre-commit.ci run

stefanv:
pre-commit.ci autofix

content/posts/numpy/numpy-rng/index.md
```diff
@@ -46,7 +46,7 @@ rng = np.random.default_rng()
...
+If you want to use a seed for reproducibility, [the NumPy documentation](https://numpy.org/doc/stable/reference/random/index.html#quick-start) recommends using a large random number. The first reason is that this increases the probability of having a different seed than anyone else and thus make sure you will have independent results. The second reason is that using only small numbers for your seeds can result into biases as they will not cover the state space of the RNG. You should use at least 128-bit numbers (check [this comment](https://github.com/numpy/numpy/issues/25778#issuecomment-1930441151) by Robert Kern for more information)[^1]. To generate a 128-bit random number for your seed you can rely on the [secrets module](https://docs.python.org/3/library/secrets.html):
```
stefanv, line 49:
```suggestion
If you want to use a seed for reproducibility, [the NumPy documentation](https://numpy.org/doc/stable/reference/random/index.html#quick-start) recommends using a large random number. The first reason is that this increases the probability of having a different seed than anyone else and thus make sure you will have independent results. The second reason is that using only small numbers for your seeds can result into biases as they will not cover the state space of the RNG. You should use at least 128-bit numbers (check [this comment](https://github.com/numpy/numpy/issues/25778#issuecomment-1930441151) by Robert Kern for more information)[^1]. To generate a 128-bit random number for your seed you can rely on the [secrets module](https://docs.python.org/3/library/secrets.html):
```

What do you mean they will not cover the state space of the PRNG? Presumably, numbers generated from a small seed has just as much a chance of accomplishing that as from large seeds.

stefanv (commented):
Thanks @albertcthomas! I requested a small clarification.

content/posts/numpy/numpy-rng/index.md
albertcthomas: Yes I thought the same at first :) It's not about the sequence of numbers you will get with one seed but about the first number you will generate with your RNG after it has been seeded. If you only seed your RNG with small seeds (smaller than the state space), and you generate an integer from it, there are for instance numbers that you will never get as the first number thus creating a bias. https://www.pcg-random.org/posts/cpp-seeding-surprises.html#bias

albertcthomas:
Thanks for the pre-commit fix. I also checked the rendering which looks good even though the footnotes in the raw file looks strange so we're good.

content/posts/numpy/numpy-rng/index.md
albertcthomas: I could add a small explanation. or maybe the link with Robert Kern's comment appears a bit too far.

albertcthomas: I added some details here.

content/posts/numpy/numpy-rng/index.md
```diff
@@ -46,7 +46,7 @@ rng = np.random.default_rng()
...
+If you want to use a seed for reproducibility, [the NumPy documentation](https://numpy.org/doc/stable/reference/random/index.html#quick-start) recommends using a large random number, where large means at least 128 bits. The first reason for using a large random number is that this increases the probability of having a different seed than anyone else and thus independent results. The second reason is that relying only on small numbers for your seeds can lead to biases as they do not fully explore the state space of the RNG. This limitation implies that the first number generated by your RNG may not seem as random as expected due to inaccessible first internal states. For example, some numbers will never be produced as the first output. One possibility would be to pick the seed at random in the state space of the RNG but [according to Robert Kern](https://github.com/numpy/numpy/issues/25778#issuecomment-1930441151) a 128-bit random number is large enough[^1]. To generate a 128-bit random number for your seed you can rely on the [secrets module](https://docs.python.org/3/library/secrets.html):
```
albertcthomas, line 49: I added some details here following your question @stefanv

stefanv: I understand now; if the seed is 32-bit, it can never generate all starting values for a 128-bit space.

stefanv (approved):

jarrodmillman (approved):
Thanks!!

content/posts/numpy/numpy-rng/index.md
albertcthomas: Yes exactly

albertcthomas:
Thanks again for the reviews :)

stefanv:
Thanks, Albert! You may have seen that we've enabled comments on the blog. It's a bit late for yours, since comments came in via issues, but should elicit some discussion on future posts.

albertcthomas:
Ah this is great. If I may (only a suggestion), it would be nice for the authors to be able to add a link to their names at the top of the posts redirecting to the url of their choice (personal page, twitter/X account, GitHub profile, …).

PR #193, opened 2024-02-25
Add author link to numpy RNG post

albertcthomas:
Adding author info to Numpy RNG blog post following  https://github.com/scientific-python/scientific-python-hugo-theme/pull/503

Thanks!

### DIFF
--- a/content/authors/albertcthomas.md
+++ b/content/authors/albertcthomas.md
@@ -5,4 +5,4 @@ title: Albert Thomas
 👋 Hi!
 
 - My website is https://albertcthomas.github.io/
-- You can also find me on X/twitter at https://twitter.com/albertcthomas
+- You can also find me on X/Twitter at https://twitter.com/albertcthomas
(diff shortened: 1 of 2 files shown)


tupui:
Thanks @albertcthomas, we just need to wait a bit as the theme is not yet synchronised there. We are on v0.11 and we will wait for v0.12.

@jarrodmillman might be able to give a timeline (or as I suspect he might do, make it happen now 😅)

albertcthomas:
Sure, no rush this is not urgent. I just did the PR now while it was on my mind :)

jarrodmillman (approved):

albertcthomas:
Thanks a lot @jarrodmillman

albertcthomas:
Yes this is perfect thanks. At some point if you need something that requires a bit less maintenance, just one page with all the authors and link to the url of their choice is also perfectly fine but the current solution is very nice!

PR #197, opened 2024-03-07
Clean up URL handling in comments partial

stefanv:


stefanv:
Merging to see if this works correctly. Can only test when deployed to blog.sp.org.

PR #206, opened 2024-03-22
Fix aTSP math rendering

mjschwenne:
Currently, all of the math rendering is broken on the asymmetric traveling salesperson posts. I believe this is due to using `\[`, `\]`, `\(` and `\)` for math delimitiers. I've updated them to use `$` and `$$`.

image


tupui (approved):
Thanks!

PR #211, opened 2024-04-10
Disable relative URLs

jarrodmillman:
Not sure why we enabled this. But my understanding is this is not recommended.

It states here https://gohugo.io/content-management/urls/#relative-urls
```
Do not enable this option unless you are creating a serverless site, navigable via the file system.
```

- [ ] The main subject relates to at least one project affiliated to the Scientific Python Ecosystem.
- [ ] I have the right to publish the content under BSD 3-Clause License for the code and Creative Common CC-BY-4.0 License for the text.
- [ ] Images have been compressed using a tool like `pngquant`.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -2,7 +2,6 @@ baseURL: "https://blog.scientific-python.org/"
 languageCode: "en-us"
 title: "Blog"
 theme: scientific-python-hugo-theme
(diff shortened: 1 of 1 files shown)


tupui (approved):
We can try

Issue #158, opened 2023-10-31
Fix breadcrumb


jarrodmillman:
![2023-10-30T21:05:16,302124489-07:00](https://github.com/scientific-python/blog.scientific-python.org/assets/123428/d9b535ce-b36b-4a53-8ae5-3a52af275aa6)

See `Posts > > ....`

jarrodmillman:
It has a few more to fix.

stefanv:
This looks like it's fixed: https://blog.scientific-python.org/scientific-python/dev-summit-1-development-guide/

agriyakhetarpal:
It has been fixed for a while indeed; just checked this blog post in specific and a few other posts. Closing!

PR #215, opened 2024-04-12
Add missing title (#158)

alphapapa:
See <https://github.com/scientific-python/blog.scientific-python.org/issues/158>.


tupui (approved):
Thanks!

PR #236, opened 2024-09-16
Do not ignore errors during build

stefanv:


jarrodmillman (approved):

bsipocz (approved):
I think this should go in.

PR #237, opened 2024-09-16
0.134 summaries

stefanv:
Ready blog for 0.134.1 upgrade

See https://github.com/scientific-python/scientific-python-hugo-theme/pull/640


jarrodmillman (approved):
Thanks!

PR #251, opened 2025-05-20
Add summary to pypalettes post

stefanv:
And thereby fix the first heading rendering on the index page.


stefanv:
@scientific-python/blog-editors please take a look; our blog front page is impacted

